PR #17, opened 2024-04-12
Relax astrapy dependency requirement to accept v 1.*

hemidactylus:


libs/astradb/Makefile
```diff
@@ -58,7 +58,7 @@ check_imports: $(shell find langchain_astradb -name '*.py')
 
 help:
 	@echo '----'
-	@echo 'check_imports				- check imports'
+	@echo 'check_imports				        - check imports'
```
cbornet, line 61: Still doesn't seem aligned

cbornet (approved):

nicoloboschi (approved):
LGTM
I just wonder if we want to add integration tests forcing the installation to 0.7.x to ensure we keep it compatible "for real"

PR #21, opened 2024-04-16
set astrapy version to ^1

jordanrfrazier:
Updates astrapy min version to ^1, which will allow us to use new features of astrapy for `vectorize` work.

### DIFF
--- a/libs/astradb/tests/integration_tests/test_storage.py
+++ b/libs/astradb/tests/integration_tests/test_storage.py
@@ -21,8 +21,6 @@ def _has_env_vars() -> bool:
 
 @pytest.fixture
 def astra_db() -> AstraDB:
-    from astrapy.db import AstraDB
-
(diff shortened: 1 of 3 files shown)


libs/astradb/tests/integration_tests/test_storage.py
```diff
@@ -21,8 +21,6 @@ def _has_env_vars() -> bool:
 
 @pytest.fixture
 def astra_db() -> AstraDB:
-    from astrapy.db import AstraDB
```
jordanrfrazier, line 24: These imports already exist above, not sure if they serve another purpose here

hemidactylus: good catch! Probably a relic from the time in langchain_community, before a pyproject.toml brought order in dependencies :)

hemidactylus (approved):

PR #22, opened 2024-04-17
feat: add support in vector_store for server-side embeddings 

jordanrfrazier:
Adds the `$vectorize` functionality to the vector store. Still uses the legacy astrapy paths.

Note: does not add functionality to the cache.

### DIFF
--- a/libs/astradb/langchain_astradb/vectorstores.py
+++ b/libs/astradb/langchain_astradb/vectorstores.py
@@ -272,11 +272,12 @@ class AstraDBVectorStore(VectorStore):
         # "vector-related" settings
         self.metric = metric
         embedding_dimension: Union[int, Awaitable[int], None] = None
-        if setup_mode == SetupMode.ASYNC:
-            embedding_dimension = self._aget_embedding_dimension()
-        elif setup_mode == SetupMode.SYNC or setup_mode == SetupMode.OFF:
-            embedding_dimension = self._get_embedding_dimension()
-
+        if self.embedding is not None:
+            if setup_mode == SetupMode.ASYNC:
+                embedding_dimension = self._aget_embedding_dimension()
+            elif setup_mode == SetupMode.SYNC or setup_mode == SetupMode.OFF:
+                embedding_dimension = self._get_embedding_dimension()
+    
         # indexing policy setting
         self.indexing_policy: Dict[str, Any] = self._normalize_metadata_indexing_policy(
             metadata_indexing_include=metadata_indexing_include,
--- a/libs/astradb/langchain_astradb/vectorstores.py
+++ b/libs/astradb/langchain_astradb/vectorstores.py
@@ -181,7 +181,7 @@ class AstraDBVectorStore(VectorStore):
         Args:
             embedding: the embeddings function or service to use.
                 This enables client-side embedding functions or calls to external
-                embedding providers. Only one of `embedding` or 
+                embedding providers. Only one of `embedding` or
                 `collection_vector_service_options` can be provided.
             collection_name: name of the Astra DB collection to create/use.
             token: API token for Astra DB usage.
@@ -216,9 +216,9 @@ class AstraDBVectorStore(VectorStore):
                 This dict must conform to to the API specifications
                 (see docs.datastax.com/en/astra/astra-db-vector/api-reference/
                 data-api-commands.html#advanced-feature-indexing-clause-on-createcollection)
-            collection_vector_service_options: specifies the use of server-side 
-                embeddings within Astra DB. Only one of `embedding` or 
-                `collection_vector_service_options` can be provided. 
+            collection_vector_service_options: specifies the use of server-side
+                embeddings within Astra DB. Only one of `embedding` or
+                `collection_vector_service_options` can be provided.
                 NOTE: this feature is currently in beta.
 
         Note:
@@ -1541,7 +1541,7 @@ class AstraDBVectorStore(VectorStore):
             metadatas: metadata dicts for the texts.
             ids: ids to associate to the texts.
             collection_vector_service_options: specifies the use of server-side
-                embeddings within Astra DB. Only one of `embedding` or 
+                embeddings within Astra DB. Only one of `embedding` or
                 `collection_vector_service_options` can be provided.
             **kwargs: you can pass any argument that you would
                 to :meth:`~add_texts` and/or to the 'AstraDBVectorStore' constructor
--- a/libs/astradb/langchain_astradb/vectorstores.py
+++ b/libs/astradb/langchain_astradb/vectorstores.py
@@ -219,7 +219,7 @@ class AstraDBVectorStore(VectorStore):
                 data-api-commands.html#advanced-feature-indexing-clause-on-createcollection)
             collection_vector_service_options: specifies the use of server-side embeddings
                 within Astra DB. Only one of `embedding` or `collection_vector_service_options`
-                can be provided.
+                can be provided. NOTE: this feature is currently in beta.
 
         Note:
             For concurrency in synchronous :meth:`~add_texts`:, as a rule of thumb, on a
@@ -517,7 +517,6 @@ class AstraDBVectorStore(VectorStore):
         #
         documents_to_insert = [
             {
-                "content": b_txt,
                 "_id": b_id,
                 "$vectorize": b_txt,
                 "metadata": b_md,
--- a/libs/astradb/langchain_astradb/vectorstores.py
+++ b/libs/astradb/langchain_astradb/vectorstores.py
@@ -219,7 +219,8 @@ class AstraDBVectorStore(VectorStore):
             collection_vector_service_options: specifies the use of server-side
                 embeddings within Astra DB. Only one of `embedding` or
                 `collection_vector_service_options` can be provided.
-                NOTE: this feature is currently in beta.
+                NOTE: This feature is under current development.
+
 
         Note:
             For concurrency in synchronous :meth:`~add_texts`:, as a rule of thumb, on a
@@ -326,15 +327,16 @@ class AstraDBVectorStore(VectorStore):
         return self.embedding_dimension
 
     @property
-    def embeddings(self) -> Embeddings:
-        if self.collection_vector_service_options is not None:
-            raise ValueError(
-                "Server-side embeddings are in use, no client-side embeddings\
-                    available."
-            )
-
-        assert self.embedding is not None
+    def embeddings(self) -> Optional[Embeddings]:
+        """
+        Accesses the supplied embeddings object. If using server-side embeddings,
+        this will return None.
+        """
         return self.embedding
+
+    def _using_vectorize(self) -> bool:
+        """Indicates whether server-side embeddings are being used."""
+        return self._using_vectorize()
 
     def _select_relevance_score_fn(self) -> Callable[[float], float]:
         """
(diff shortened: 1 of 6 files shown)


libs/astradb/tests/integration_tests/test_vectorstores.py
```diff
@@ -383,12 +477,56 @@ async def test_astradb_vectorstore_from_x_async(
...
+            ].page_content == "HeeH"
+        finally:
+            if not SKIP_COLLECTION_DELETE:
+                await v_store_4.adelete_collection()
+            else:
+                await v_store_4.aclear()
+
+    @pytest.mark.parametrize(
+        "vector_store", ["store_someemb", "vectorize_store"]
+    )
```
jordanrfrazier, line 522: Thoughts on if this should be replicated on all applicable tests? e.g. do we need to run `test_astradb_vectorstore_similarity_scale` for the vectorize store as well? 

I think it would be good to have, but don't want to go overboard on vectorize yet.

libs/astradb/tests/integration_tests/test_vectorstores.py
```diff
@@ -111,6 +123,9 @@ def _has_env_vars() -> bool:
 
 
 @pytest.fixture(scope="session")
+@pytest.mark.skipif(
+    is_vector_service_available(), reason="only run vectorize-specific tests in dev"
+)
```
jordanrfrazier, line 128: I don't love this, but unsure of a better alternative to avoid running the rest of the tests in vectorize. I didn't want to make a new test suite with duplicated tests either.

libs/astradb/tests/integration_tests/test_vectorstores.py
```diff
@@ -383,12 +498,58 @@ async def test_astradb_vectorstore_from_x_async(
...
+                0
+            ].page_content == "HeeH"
+        finally:
+            if not SKIP_COLLECTION_DELETE:
+                await v_store_4.adelete_collection()
+            else:
+                await v_store_4.aclear()
+
+    # @pytest.mark.parametrize("vector_store", ["store_someemb", "vectorize_store"])
+    @pytest.mark.parametrize("vector_store", ["vectorize_store"])
```
jordanrfrazier, line 542: TODO: Fix. However, issue I'm stuck on here is that the `score` for sim search is coming up ~0.85 -- unsure why we're not seeing close to 1.

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -209,6 +216,10 @@ def __init__(
                 This dict must conform to to the API specifications
                 (see docs.datastax.com/en/astra/astra-db-vector/api-reference/
                 data-api-commands.html#advanced-feature-indexing-clause-on-createcollection)
+            collection_vector_service_options: specifies the use of server-side
+                embeddings within Astra DB. Only one of `embedding` or
+                `collection_vector_service_options` can be provided.
+                NOTE: this feature is currently in beta.
```
nicoloboschi, line 222: do we really need to state this here? This will become obsolete and it's not good-looking for users

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -295,6 +327,13 @@ async def _aget_embedding_dimension(self) -> int:
 
     @property
     def embeddings(self) -> Embeddings:
+        if self.collection_vector_service_options is not None:
+            raise ValueError(
+                "Server-side embeddings are in use, no client-side embeddings\
```
nicoloboschi, line 332: I'd point out that `collection_vector_service_options ` has been passed and that's why you can't get the embeddings client

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -783,6 +868,81 @@ async def asimilarity_search_with_score_id_by_vector(
             )
         ]
 
+    def _similarity_search_with_score_id_with_vectorize(
```
nicoloboschi, line 871: can we have a single private method that check wheter to use vectorize or not? It should be easier to maintain

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -783,6 +868,81 @@ async def asimilarity_search_with_score_id_by_vector(
...
+                    page_content=hit["$vectorize"],
+                    metadata=hit["metadata"],
+                ),
+                hit["$similarity"],
+                hit["_id"],
+            )
+            for hit in hits
+        ]
+
+    async def _asimilarity_search_with_score_id_with_vectorize(
```
nicoloboschi, line 911: can we have a single private method that check wheter to use vectorize or not? It should be easier to maintain

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -799,12 +959,24 @@ def similarity_search_with_score_id(
...
+            )
+        elif self.embedding is not None:
+            embedding_vector = self.embedding.embed_query(query)
+            return self.similarity_search_with_score_id_by_vector(
+                embedding=embedding_vector,
+                k=k,
+                filter=filter,
+            )
+        else:
+            raise ValueError(
```
nicoloboschi, line 976: I think this shouldn't happen, right? 
I wonder if it's better to just add the assertion

nicoloboschi (dismissed):
Great work!!
I've left some comments

libs/astradb/langchain_astradb/vectorstores.py
jordanrfrazier: Astrapy has a similar comment for this object, though it says "This feature is under development", which sounds better. I could swap, or I remove completely.

libs/astradb/langchain_astradb/vectorstores.py
jordanrfrazier: Ah good catch. Thanks

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: It's a slightly tricky refactoring (the original paths expose _by_vector methods with the actual logic, which the vectorize path would not), but it could be advantageous on the long run.

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -135,8 +136,8 @@ def _normalize_metadata_indexing_policy(
     def __init__(
         self,
         *,
-        embedding: Embeddings,
         collection_name: str,
+        embedding: Optional[Embeddings] = None,
```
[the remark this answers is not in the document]
hemidactylus: Er... scrap that. I just did not pay attention to the wonderful `*` that comes first.
It's all good now.

libs/astradb/tests/integration_tests/test_vectorstores.py
hemidactylus: Why avoiding the rest of the tests in the first place? I thought one could manually, locally run pytest with appropriate flags (`-k ...`) for fast iteration, but preserve the full thing for the CI steps.

Come to think of it, how would that be tested in CI? This test setup seems something that can run only locally, short of altering the repo secrets or creating a second ("dev") set of creds (with an additional DB-creating layer).
To be clear, I am not necessarily advocating for such increased complexity for a temporary limitation.

libs/astradb/langchain_astradb/vectorstores.py
jordanrfrazier: Do you mean something like:
```
    def _using_vectorize(self) -> bool:
        return self.collection_vector_service_options is not None
```
and 
```
def similarity_search(
        self,
        query: str,
        k: int = 4,
        filter: Optional[Dict[str, Any]] = None,
        **kwargs: Any,
    ) -> List[Document]:
        if _using_vectorize():
            return [
                doc
                for (doc, _, _) in self._similarity_search_with_score_id_with_vectorize(
                    query,
                    k,
                    filter=filter,
                )
            ]
        else:
            embedding_vector = self.embedding.embed_query(query)
            return self.similarity_search_by_vector(
                embedding_vector,
                k,
                filter=filter,
            )
```

or are you implying a larger refactoring of some sort?

libs/astradb/langchain_astradb/vectorstores.py
jordanrfrazier: Yeah, you're right. We can assume that `self.embedding` exists if we're not using vectorize. If it fails, then an invariant has been broken prior to this call.

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -326,16 +327,17 @@ async def _aget_embedding_dimension(self) -> int:
...
+    def embeddings(self) -> Optional[Embeddings]:
+        """
+        Accesses the supplied embeddings object. If using server-side embeddings,
+        this will return None.
+        """
         return self.embedding
 
+    def _using_vectorize(self) -> bool:
+        """Indicates whether server-side embeddings are being used."""
+        return self._using_vectorize()
```
hemidactylus, line 339: I think this looks like a Ctrl-V too much, which would lead to the max-recursion-depth-exceeded error once called :)

hemidactylus:
Very nice and comprehensive work!

My only suggestion concerns the fixtures around testing. With the vectorize feature eventually landing in prod, it could be better to prepare for that (something like a twin set of fixtures, vector- and vectorize-based) and add a small amount of skipifs to momentarily block the latter in prod? For ways to combine this with the pytest parametrize, I found a nice way to condition items of a parametrize (see [here](https://github.com/datastax/astrapy/blob/3b369b4c2e229979bec995a9368d4887e644159b/tests/idiomatic/integration/test_admin.py#L45)).

libs/astradb/langchain_astradb/vectorstores.py
jordanrfrazier: sed betrayal

good catch

libs/astradb/tests/integration_tests/test_vectorstores.py
jordanrfrazier: You're right -- no reason not to run the rest of the tests. 

Removing these skips, we should be running all tests (minus vectorize) in any environment, and including vectorize tests when in dev.

jordanrfrazier:
> My only suggestion concerns the fixtures around testing. With the vectorize feature eventually landing in prod, it could be better to prepare for that (something like a twin set of fixtures, vector- and vectorize-based) and add a small amount of skipifs to momentarily block the latter in prod? For ways to combine this with the pytest parametrize, I found a nice way to condition items of a parametrize (see [here](https://github.com/datastax/astrapy/blob/3b369b4c2e229979bec995a9368d4887e644159b/tests/idiomatic/integration/test_admin.py#L45)).

Cool, this is a neat technique to know. I did look into using this pattern, but since we have several tests that use one or the other of the `StoreEmbedding` vs. `ParseEmbedding`, it was easier to just put a single `skipif` on the `vectorize_store`, which can be removed once vectorize is in prod. 

We could probably refactor the tests into separate modules for the different embeddings, then use the linked pattern in a further revision.

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: I took Nicolò's comment to mean the larger refactoring - but I am not sure it's worth the saved duplication in hindsight.

libs/astradb/tests/integration_tests/test_vectorstores.py
hemidactylus: This seems solved now in light of the prepended "query: " or "passage: " added behind the scenes, correct?

hemidactylus (approved):

PR #26, opened 2024-04-24
ci: skip integration tests during release process

nicoloboschi:


mieslep (approved):
approved, though with some reservation @nicoloboschi :)

PR #30, opened 2024-05-20
Environment variables for Astra DB init

hemidactylus:
This PR modifies the flow of the Astra DB environment utils so as to fall back to standard environment variables if Astra DB secrets/connection parameters are required and not provided explicitly.

The presence of such variables in the environment does not alter the current behaviour (except for the case of no namespace provided and one defined in the environment).

An extensive unit test probes the various init combinations.

### DIFF
--- a/libs/astradb/langchain_astradb/cache.py
+++ b/libs/astradb/langchain_astradb/cache.py
@@ -112,15 +112,18 @@ class AstraDBCache(BaseCache):
 
         Args:
             collection_name: name of the Astra DB collection to create/use.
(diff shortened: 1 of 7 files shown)


nicoloboschi (approved):

nicoloboschi (commented):
@hemidactylus shall we update the docstring to explain this chain ?

erichare:
@hemidactylus looks great!

cbornet:
@hemidactylus wouldn't it be useful to support this in astrapy ?

PR #33, opened 2024-05-25
Upgrade vectorize support to Astra's Public Preview available service

hemidactylus:
I have extended the vectorize support to cover for what has been launched in public preview this week.
Namely, this means bumping astrapy to access the latest parameters "service options", and the various ways of passing secrets to embedding services:
- through the KMS
- through a header (not through secrets stored in astra)
- no secrets (i.e. right now nVidia nemo).
A further param ( <--> the header) is added to the vector store and the helper "util" class to achieve that (plus further ValueError for new unacceptable combinations of parameters)

Much has changed in the tests, which end up unavoidably convoluted for testing these things according to what services are available on which environments: the "main" vectorize is now openai, so that "regular" astra have it. This requires
(a) an openai key - I have not entered any in the CI yet
(b) which must be also set as a secret in the KMS for that account (same)
These two things I have yet to fix on the repo CI, but so far I tested locally and everything is perfect!
The above cover for the kms+header usages.

What remains is the no-auth (=nvidia nemo) way. Right now, this is available only if your Astra org is manually flagged to try that. That is the reason I switched the standard vectorize testing service to openai. However, I have access to an org "in the club", so I could test that vectorize flavour as well: all green.


nicoloboschi (approved):
LGTM

libs/astradb/codespell_ignore_words.txt
```diff
@@ -0,0 +1 @@
+Haa
```
jordanrfrazier, line 1: What's this? I see it in the test but wouldn't expect it to complain about there

libs/astradb/langchain_astradb/utils/astradb.py
```diff
@@ -138,19 +140,29 @@ def __init__(
         collection_vector_service_options: Optional[
             CollectionVectorServiceOptions
         ] = None,
+        collection_embedding_api_key: Optional[str] = None,
```
jordanrfrazier, line 143: Why is the key added outside of the `CollectionVectorServiceOptions` class?

libs/astradb/tests/integration_tests/test_vectorstores.py
```diff
@@ -34,16 +34,60 @@
...
+    model_name="text-embedding-3-small",
+    authentication={
+        "providerKey": f"{os.environ.get('SHARED_SECRET_NAME_OPENAI', '')}.providerKey",
+    },
+)
+openAIVectorizeOptionsHeader = CollectionVectorServiceOptions(
+    provider="openai",
+    model_name="text-embedding-3-small",
+)
+nvidiaVectorizeOptions = CollectionVectorServiceOptions(
```
jordanrfrazier, line 59: nit: should these use snake case?

jordanrfrazier (approved):

libs/astradb/langchain_astradb/utils/astradb.py
hemidactylus: This is a different way to operate the embedding service: the secret is not in the KMS, rather it is passed at each data read/write call as a header.
Since this is also supported by the data API, this test tests it.

libs/astradb/codespell_ignore_words.txt
jordanrfrazier: Haha weird, why is `Huu` okay? 
```
            texts=["Haa", "Huu"],
```

libs/astradb/langchain_astradb/utils/astradb.py
jordanrfrazier: Ah, I see. I was looking at an old version of `astrapy`. The `auth` there can be used instead of `collection_embedding_api_key` here. Thanks

PR #36, opened 2024-06-15
fix: add default indexing policy to chat history

jordanrfrazier:
Adds a default indexing policy to the chat message history class. 

The reason for this is that the default policy when creating a vector store through `AstraDBVectorStore` uses this policy. Meaning, if you create a collection with the class, you cannot initialize chat history, as you'll get this error:

```
E               ValueError: Astra DB collection 'test' is detected as having the following indexing policy: {"allow": ["metadata"]}. This is incompatible with the requested indexing policy for this object. Consider reindexing anew on a fresh collection with the requested indexing policy, or alternatively align the requested indexing settings to the collection to keep using it.

```

I think it makes sense in this case to make the default indexing policy the same here. 

https://github.com/langchain-ai/langchain-datastax/blob/77d3d81d981b9dc28296d85e461b7b803737d904/libs/astradb/langchain_astradb/vectorstores.py#L57

### DIFF
--- a/libs/astradb/langchain_astradb/chat_message_histories.py
+++ b/libs/astradb/langchain_astradb/chat_message_histories.py
@@ -18,6 +18,8 @@ from langchain_astradb.utils.astradb import (
     _AstraDBCollectionEnvironment,
 )
 
+# indexing options when creating a collection
+DEFAULT_INDEXING_OPTIONS = {"allow": ["metadata"]}
 DEFAULT_COLLECTION_NAME = "langchain_message_store"
 
 
@@ -63,6 +65,7 @@ class AstraDBChatMessageHistory(BaseChatMessageHistory):
             namespace=namespace,
             setup_mode=setup_mode,
             pre_delete_collection=pre_delete_collection,
+            default_indexing_policy=DEFAULT_COLLECTION_NAME,
         )
 
         self.collection = self.astra_env.collection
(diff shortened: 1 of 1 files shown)


jordanrfrazier:
Open to discussions here - I didn't dig very deeply into this. I was writing integration tests for langflow and running into this error.

jordanrfrazier:
> A side comment: the reason the CI fails is unrelated to this PR. I encountered the same while upgrading the vectorize and came up with a fix (see #37 for a brief explanation). I suggest to bring that one in this branch once 37 is merged, to get smooth CI here as well.
> 
> Now to the main question. While tailoring the indexing policy for chat history to the use is in general a good thing (fight SAI waste!), I am not sure I get the reason why one should use the _same_ collection for both vectorstore and chat history. The format of the documents in them is different as far as I remember, no? Would there be a real-world situation (outside of tests perhaps) where a single collection might serve both purposes? Note: I might be missing some context here `¯\_(ツ)_/¯`

I could probably imagine up some use-case where we'd want metadata attached to each message, such as "time sent" for accessing only the last five minutes of chat history, or "location" for location-based recommendations in a conversation, which would reduce the number of stores to query. However, it's true the main reason was during testing I ran into this - I wanted to use a single fixture to set up my Astra collections before running tests, which used the `AstraDBVectorStore` class (thus defaulting to `{allow: metadata}`). I ended up using the chat history in a separate fixture to set up that specific class, so very easy to avoid this issue. 

I don't have a stronger argument than the above, so very happy to pass on this as "working-as-intended"

jordanrfrazier:
Closing per above comment

PR #40, opened 2024-07-09
Refactor MMR flow so as to enable it with vectorize stores

hemidactylus:
The good:

the Data API admits a "includeSortVector" flag for $vector and $vectorize searches alike, so that by getting the query vector back in the response we can enable MMR for all vector stores. (at the cost of a little refactoring of private methods. A slight increase in code convolution, sadly.)

This query vector is found outside of the list of "documents" from the API, which requires gymnastics to retrieve it (a callback, as the "core" astrapy requires). This will be simplified when switching to astrapy>1 "idiomatic" library: they expose the sort vector directly for easy access.

This PR requires astrapy>=1.3 (so I changed the pyproject.toml). I had some trouble with `poetry lock` (numpy-related unsatisfiable versions), so I went for `poetry lock --no-update` instead. This might need to be inspected closely.


hemidactylus:
@nicoloboschi @jordanrfrazier FYI

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1283,9 +1283,107 @@ async def asimilarity_search_with_score(
...
+        metadata_parameter: Dict[str, Any],
+        **kwargs: Any,
+    ) -> List[Document]:
+        query_vector_l: List[Optional[List[float]]] = [None]
+
+        def _response_setter(
+            resp: Dict[str, Any],
+            qvl: List[Optional[List[float]]] = query_vector_l,
+        ) -> None:
+            qvl[0] = resp["status"]["sortVector"]
```
jordanrfrazier, line 1301: This is pretty cool 💪

jordanrfrazier (approved):

nicoloboschi (approved):

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: you mean the "dirtiest of dirty tricks to achieve pass-by-reference" ? LOL

PR #42, opened 2024-07-23
Full porting to astrapy 1.3+

hemidactylus:
This is a large rewrite of the internals in order to completely migrate to using the recent, idiomatic astrapy.

### Motivation

- enabling non-Astra Data API (such as HCD. I could run all integration testing on HCD as well)
- simplify most of the Data API newer features (such as $vectorize)
- use an astrapy API that is extensively documented and actively maintained

### Guide to the PR

All the newest astrapy things are now usable (e.g. environments, TokenProvider, EmbeddingHeaderProvider classes and so on).

Along the way, several improvements came along:

- deprecating (but still 100% working) passing "clients" to the constructors. Preferred: secret/strings in init methods
- issuing warning if "legacy" parameters are used (as part of the full alignment to the modern astrapy). But without  breaking changes. Everything is 100% compatible with previous usage patterns
- reorganization/modernization of the whole test suite (fixture re-use, uniform test function scheme, etc)
- more testing added, esp. thoroughly probing the (totally rewritten) insert-very-many-with-possible-overwrite flows (sync/async)

### Notes

1. The internals have changed a lot. Hence, code that directly uses the internals of these classes (e.g. using `vector_store.collection`) would stop working. Such usage, anyway, is brittle in the first place and should not be encouraged.
2. The document loader maintains the same signature, including its `nb_prefetched` parameter, which however is ignored - actually a warning is emitted if a prefetch is requested. This reflects the fact that prefetched finds have been found to be problematic in some corner cases on AstraPy and are at the moment disabled. Functionally, this has no effect on the document loader class.

### DIFF
--- a/libs/astradb/langchain_astradb/cache.py
+++ b/libs/astradb/langchain_astradb/cache.py
@@ -3,6 +3,7 @@ import json
 from functools import lru_cache, wraps
 from typing import Any, Awaitable, Callable, Generator, List, Optional, Tuple, Union
 
+from astrapy.authentication import TokenProvider
 from astrapy.db import AstraDB, AsyncAstraDB, logger
 from langchain_core.caches import RETURN_VAL_TYPE, BaseCache
 from langchain_core.embeddings import Embeddings
@@ -93,7 +94,7 @@ class AstraDBCache(BaseCache):
         self,
         *,
         collection_name: str = ASTRA_DB_CACHE_DEFAULT_COLLECTION_NAME,
-        token: Optional[str] = None,
+        token: Optional[Union[str, TokenProvider]] = None,
         api_endpoint: Optional[str] = None,
         environment: Optional[str] = None,
         astra_db_client: Optional[AstraDB] = None,
@@ -113,8 +114,10 @@ class AstraDBCache(BaseCache):
 
         Args:
             collection_name: name of the Astra DB collection to create/use.
-            token: API token for Astra DB usage. If not provided, the environment
-                variable ASTRA_DB_APPLICATION_TOKEN is inspected.
+            token: API token for Astra DB usage, either in the form of a string
+                or a subclass of `astrapy.authentication.TokenProvider`.
+                If not provided, the environment variable
+                ASTRA_DB_APPLICATION_TOKEN is inspected.
             api_endpoint: full URL to the API endpoint, such as
                 `https://<DB-ID>-us-east1.apps.astra.datastax.com`. If not provided,
                 the environment variable ASTRA_DB_API_ENDPOINT is inspected.
@@ -298,7 +301,7 @@ class AstraDBSemanticCache(BaseCache):
         self,
         *,
         collection_name: str = ASTRA_DB_SEMANTIC_CACHE_DEFAULT_COLLECTION_NAME,
-        token: Optional[str] = None,
+        token: Optional[Union[str, TokenProvider]] = None,
         api_endpoint: Optional[str] = None,
         environment: Optional[str] = None,
         astra_db_client: Optional[AstraDB] = None,
@@ -324,8 +327,10 @@ class AstraDBSemanticCache(BaseCache):
 
         Args:
             collection_name: name of the Astra DB collection to create/use.
-            token: API token for Astra DB usage. If not provided, the environment
-                variable ASTRA_DB_APPLICATION_TOKEN is inspected.
+            token: API token for Astra DB usage, either in the form of a string
+                or a subclass of `astrapy.authentication.TokenProvider`.
+                If not provided, the environment variable
+                ASTRA_DB_APPLICATION_TOKEN is inspected.
             api_endpoint: full URL to the API endpoint, such as
                 `https://<DB-ID>-us-east1.apps.astra.datastax.com`. If not provided,
                 the environment variable ASTRA_DB_API_ENDPOINT is inspected.
--- a/libs/astradb/langchain_astradb/cache.py
+++ b/libs/astradb/langchain_astradb/cache.py
@@ -95,6 +95,7 @@ class AstraDBCache(BaseCache):
         collection_name: str = ASTRA_DB_CACHE_DEFAULT_COLLECTION_NAME,
         token: Optional[str] = None,
         api_endpoint: Optional[str] = None,
+        environment: Optional[str] = None,
         astra_db_client: Optional[AstraDB] = None,
         async_astra_db_client: Optional[AsyncAstraDB] = None,
         namespace: Optional[str] = None,
@@ -117,10 +118,19 @@ class AstraDBCache(BaseCache):
             api_endpoint: full URL to the API endpoint, such as
                 `https://<DB-ID>-us-east1.apps.astra.datastax.com`. If not provided,
                 the environment variable ASTRA_DB_API_ENDPOINT is inspected.
-            astra_db_client: *alternative to token+api_endpoint*,
-                you can pass an already-created 'astrapy.db.AstraDB' instance.
-            async_astra_db_client: *alternative to token+api_endpoint*,
-                you can pass an already-created 'astrapy.db.AsyncAstraDB' instance.
+            environment: a string specifying the environment of the target Data API.
+                If omitted, defaults to "prod" (Astra DB production).
+                Other values are in `astrapy.constants.Environment` enum class.
+            astra_db_client:
+                *DEPRECATED starting from version 0.3.5.*
+                *Please use 'token', 'api_endpoint' and optionally 'environment'.*
+                you can pass an already-created 'astrapy.db.AstraDB' instance
+                (alternatively to 'token', 'api_endpoint' and 'environment').
+            async_astra_db_client:
+                *DEPRECATED starting from version 0.3.5.*
+                *Please use 'token', 'api_endpoint' and optionally 'environment'.*
+                you can pass an already-created 'astrapy.db.AsyncAstraDB' instance
+                (alternatively to 'token', 'api_endpoint' and 'environment').
             namespace: namespace (aka keyspace) where the collection is created.
                 If not provided, the environment variable ASTRA_DB_KEYSPACE is
                 inspected. Defaults to the database's "default namespace".
@@ -134,6 +144,7 @@ class AstraDBCache(BaseCache):
             collection_name=collection_name,
             token=token,
             api_endpoint=api_endpoint,
+            environment=environment,
             astra_db_client=astra_db_client,
             async_astra_db_client=async_astra_db_client,
             namespace=namespace,
@@ -153,7 +164,7 @@ class AstraDBCache(BaseCache):
             projection={
                 "body_blob": 1,
             },
-        )["data"]["document"]
+        )
         return _loads_generations(item["body_blob"]) if item is not None else None
 
     async def alookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYPE]:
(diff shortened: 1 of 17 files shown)


libs/astradb/langchain_astradb/document_loaders.py
```diff
@@ -31,46 +34,64 @@ def __init__(
...
+        environment: Optional[str] = None,
         astra_db_client: Optional[AstraDB] = None,
         async_astra_db_client: Optional[AsyncAstraDB] = None,
         namespace: Optional[str] = None,
         filter_criteria: Optional[Dict[str, Any]] = None,
         projection: Optional[Dict[str, Any]] = _NOT_SET,  # type: ignore[assignment]
         find_options: Optional[Dict[str, Any]] = None,
-        nb_prefetched: int = 1000,
+        limit: Optional[int] = None,
+        nb_prefetched: int = _NOT_SET,  # type: ignore[assignment]
```
cbornet, line 47: Use `-1` for easier typing ?

libs/astradb/langchain_astradb/document_loaders.py
```diff
@@ -81,13 +102,54 @@ def __init__(
...
+                        "dictionary is deprecated starting from version 0.3.5. "
+                        "Please switch to passing 'limit=<number>' "
+                        "directly in the constructor."
+                    ),
+                    DeprecationWarning,
+                )
+                _limit = (find_options or {})["limit"]
+        else:
+            _limit = limit
+        self.limit = _limit
```
cbornet, line 136: How about simplifying like this :
```python
_find_options = find_options.copy() if find_options else {}
if "limit" in _find_options:
   # raise and warn but don't set _limit 
   ...
self.limit = _find_options.pop("limit", limit)
if _find_options:
    # no need to get _other_option_keys
    warnings.warn(...)
```

libs/astradb/langchain_astradb/document_loaders.py
```diff
@@ -81,13 +102,54 @@ def __init__(
...
+                (
+                    "Parameter 'nb_prefetched' is not supported by the Data API "
+                    "client and will be ignored in reading document."
+                ),
+                UserWarning,
+            )
+
+        # normalizing limit and options and deprecations
+        _limit: Optional[int]
+        if "limit" in (find_options or {}):
```
cbornet, line 117: How about simplifying like this :
```python
_find_options = find_options.copy() if find_options else {}
if "limit" in _find_options:
   # raise and warn but don't set _limit 
   ...
self.limit = _find_options.pop("limit", limit)
if _find_options:
    # no need to get _other_option_keys
    warnings.warn(...)
```

libs/astradb/langchain_astradb/storage.py
```diff
@@ -76,15 +85,113 @@ async def amget(self, keys: Sequence[str]) -> List[Optional[V]]:
...
+
+        # if necessary, replace docs for the non-inserted ids
+        if ids_to_replace:
+            documents_to_replace = [
+                document
+                for document in documents_to_insert
+                if document["_id"] in ids_to_replace
+            ]
+
+            sem = asyncio.Semaphore(REPLACE_DOCUMENTS_MAX_THREADS)
```
cbornet, line 170: We use REPLACE_DOCUMENTS_MAX_THREADS  but it's not exactly threads ?

libs/astradb/langchain_astradb/utils/astradb.py
```diff
@@ -11,15 +11,25 @@
...
 TOKEN_ENV_VAR = "ASTRA_DB_APPLICATION_TOKEN"
 API_ENDPOINT_ENV_VAR = "ASTRA_DB_API_ENDPOINT"
 NAMESPACE_ENV_VAR = "ASTRA_DB_KEYSPACE"
 
-DEFAULT_VECTORIZE_SECRET_HEADER = "x-embedding-api-key"
+# Default settings for API data operations (concurrency & similar):
+# Chunk size for many-document insertions (None meaning defer to astrapy):
+DEFAULT_DOCUMENT_CHUNK_SIZE = None
+# thread/coroutine count for bulk inserts
+INSERT_DOCUMENT_MAX_THREADS = 20
```
cbornet, line 28: Maybe MAX_CONCURRENT_DOCUMENT_INSERT ?

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1610,31 +1611,7 @@ def _from_kwargs(
...
-            metric=metric,
-            batch_size=kwargs.get("batch_size"),
-            bulk_insert_batch_concurrency=kwargs.get("bulk_insert_batch_concurrency"),
-            bulk_insert_overwrite_concurrency=kwargs.get(
-                "bulk_insert_overwrite_concurrency"
-            ),
-            bulk_delete_concurrency=kwargs.get("bulk_delete_concurrency"),
-            collection_vector_service_options=collection_vector_service_options,
-        )
+        return cls(**kwargs)
```
cbornet, line 1614: I think this will crash if we get an unknown argument

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1643,24 +1620,14 @@ def from_texts(
...
-        collection_vector_service_options: Optional[
-            CollectionVectorServiceOptions
-        ] = None,
         **kwargs: Any,
     ) -> AstraDBVectorStore:
         """Create an Astra DB vectorstore from raw texts.
 
         Args:
             texts: the texts to insert.
-            embedding: the embedding function to use in the store.
```
cbornet, line 1655: Why remove ?

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1669,18 +1636,22 @@ def from_texts(
         Returns:
             an `AstraDBVectorStore` vectorstore.
         """
+        _method_args = {
```
cbornet, line 1639: Maybe it would be simpler to expose these params in the signature ?

cbornet:
LGTM, great work !
A few nits.

cbornet (approved):

libs/astradb/langchain_astradb/cache.py
```diff
@@ -112,15 +114,26 @@ def __init__(
...
                 `https://<DB-ID>-us-east1.apps.astra.datastax.com`. If not provided,
                 the environment variable ASTRA_DB_API_ENDPOINT is inspected.
-            astra_db_client: *alternative to token+api_endpoint*,
-                you can pass an already-created 'astrapy.db.AstraDB' instance.
-            async_astra_db_client: *alternative to token+api_endpoint*,
-                you can pass an already-created 'astrapy.db.AsyncAstraDB' instance.
+            environment: a string specifying the environment of the target Data API.
+                If omitted, defaults to "prod" (Astra DB production).
+                Other values are in `astrapy.constants.Environment` enum class.
+            astra_db_client:
```
cbornet, line 127: Will we give the possibility to pass/share Database/AsyncDatabase instances in the future ?

libs/astradb/langchain_astradb/document_loaders.py
hemidactylus: No strong opinion here - but the `_not_set` special object is something we've been using elsewhere in the repo. Oh, well.

libs/astradb/langchain_astradb/document_loaders.py
hemidactylus: But I wanted to raise an error if there `"limit"` in find_options _and_ an explicit `limit` at the same time, which would call for more logic in place of the oneliner `self.limit = _find_options.pop("limit", limit)`. Complexity creeps in again...

libs/astradb/langchain_astradb/storage.py
hemidactylus: Correct, well spotted. Let us rename => `MAX_CONCURRENT_DOCUMENT_REPLACEMENTS`, agnostic enough.

libs/astradb/langchain_astradb/utils/astradb.py
hemidactylus: `MAX_CONCURRENT_DOCUMENT_INSERTIONS` be it (analogy with the previous)

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: That is sort of the intention - i.e. so that faulty invocations of `_from_texts` and `_afrom_texts` behave as if the caller were calling the (actual) constructor. Including the penalties for passing unknown arguments.
Would you prefer to be more prescriptive here?

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: I assumed it was covered by the "anything you'd pass to add_text and/or the constructor" part of the docstring for `kwargs`. But I'll reinstate it, no particular opinion here.

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: This chain of non-constructor entry point that internally end up calling the constructor proved to be hard to maintain. A lot of arguments were forgotten and these various "constructor aliases" ended up being out of sync.
This was proposed as a solution to free us from the burden of worrying too much.

libs/astradb/langchain_astradb/cache.py
hemidactylus: If the needs arises, I don't see why not. I had to clearly draw a deprecation boundary around passing the "core" (aka legacy) classes and did not want to introduce further init complexity yet, but now the astrapy API is stable so why not. Let me file an issue (but keeping it separate from this PR)

libs/astradb/langchain_astradb/document_loaders.py
cbornet: Yes, that's what I was referring to in 
```python
# raise and warn but don't set _limit
```
I removed the code only to limit the number of lines but the idea is to keep the section untouched.

So the complete code:
```python
        _find_options = find_options.copy() if find_options else {}
        if "limit" in _find_options:
            if limit is not None:
                raise ValueError(
                    "Duplicate 'limit' directive supplied. Please remove it "
                    "from the 'find_options' map parameter."
                )
            else:
                warnings.warn(
                    (
                        "Passing 'limit' as part of the 'find_options' "
                        "dictionary is deprecated starting from version 0.3.5. "
                        "Please switch to passing 'limit=<number>' "
                        "directly in the constructor."
                    ),
                    DeprecationWarning,
                )
        self.limit = _find_options.pop("limit", limit)
        if _find_options:
            # no need to get _other_option_keys
            warnings.warn(...)
```

libs/astradb/langchain_astradb/vectorstores.py
cbornet: The current behavior is to ignore unknown parameters and that's also what the warning log says.
We can change to reject completely. Then the log would probably not be needed anymore.

libs/astradb/langchain_astradb/vectorstores.py
cbornet: There will still be maintenance burden of updating the `_method_args ` dict...

libs/astradb/langchain_astradb/document_loaders.py
hemidactylus: My bad, I had not seen the "_raise and warn_" for what it meant on first reading. Sure, I have adopted your proposal 100%, thank you!

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: Ah, very well spotted! I just reintroduced the "warn but proceed" behaviour for the sake of maintaining current behaviour (for as little a corner case this might be...).

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: Um, I thought I had automated that part as well using `inspect`. Now it is - hopefully no "forgetting to manually propagate new parameters" anymore.

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: Also added tests specific to this

PR #46, opened 2024-08-20
Bump ruff and update Makefile

cbornet:


hemidactylus (approved):
LGTM, thank you for doing this!

PR #51, opened 2024-08-22
Fix more ruff checks

cbornet:


libs/astradb/langchain_astradb/cache.py
```diff
@@ -75,16 +77,17 @@ def _loads_generations(generations_str: str) -> RETURN_VAL_TYPE | None:
...
-        logger.warning(
-            f"Legacy 'Generation' cached blob encountered: '{generations_str}'"
-        )
-        return generations
     except (json.JSONDecodeError, TypeError):
         logger.warning(
             f"Malformed/unparsable cached blob encountered: '{generations_str}'"
         )
         return None
+    else:
```
cbornet, line 85: see https://docs.astral.sh/ruff/rules/try-consider-else/

libs/astradb/langchain_astradb/document_loaders.py
```diff
@@ -113,6 +113,7 @@ def __init__(
                     "client and will be ignored in reading document."
                 ),
                 UserWarning,
+                stacklevel=2,
```
cbornet, line 116: see https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/

hemidactylus (approved):

PR #52, opened 2024-08-22
Vector store, refactor encoding to an Astra document

hemidactylus:
This PR factors the translations to an from the document stored in the Astra DB collection.
To do so, a new `VSDocumentEncoder` interface is proposed, providing operations such as `Document`-to-Astra and vice-versa, knowledge of the proper projection clauses when querying, and also the knowledge of whether embeddings are server-side ($vectorize) or not.

The main advantage of this refactoring lie in the next two planned phases after this PR:

1. a rethinking of the many `similarity_search*` methods, aimed at sharing as much as possible between the vectorize- and the nonvectorize- paths;
2. most important, an "auto-detect" mode for creating the Vector Store that creates a suitable VSDocumentEncoder based on the nature of the (already-existing) table.


libs/astradb/langchain_astradb/utils/encoders.py
```diff
@@ -0,0 +1,191 @@
...
+    server_side_embeddings: bool
+    content_field: str
+    base_projection: dict[str, bool]
+    full_projection: dict[str, bool]
+
+    @abstractmethod
+    def encode(
+        self,
+        content: str,
+        id: str,  # noqa: A002
```
cbornet, line 51: Maybe use `doc_id` or `document_id` instead of silencing `A002` ?

libs/astradb/langchain_astradb/utils/encoders.py
```diff
@@ -0,0 +1,191 @@
...
+from typing import Any
+
+from langchain_core.documents import Document
+from typing_extensions import override
+
+
+def _default_filter_encoder(filter_dict: dict[str, Any]) -> dict[str, Any]:
+    metadata_filter = {}
+    for k, v in filter_dict.items():
+        if k and k[0] == "$":
```
jordanrfrazier, line 13: Can you document the significance of checking for `$`? 

Is it specifically for handling astra keywords, such as `$vectorize`? Are users prohibited from specifying leading `$` in their filters; what happens if they were to do so?

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1411,7 +1341,7 @@ async def asimilarity_search_with_score(
         Returns:
             The list of (Document, score), the most similar to the query vector.
         """
-        if self._using_vectorize():
+        if self.document_encoder.server_side_embeddings:
```
jordanrfrazier, line 1344: I wonder if these if statements can now be simplified

jordanrfrazier (approved):
Looks awesome

libs/astradb/langchain_astradb/utils/encoders.py
hemidactylus: Good catch, thanks!

libs/astradb/langchain_astradb/utils/encoders.py
hemidactylus: (this part was unchanged from before the refactoring, just moved)

The reason is to not process special, Astra-specific parts of the filter, yes, but not `$vectorize` (which should not be passed in the filter ever).
But one may want to use the operators (`$or`, `$and`, etc) and we do not want to prepend them with `metadata.`.

So you get the _field names_ correctly embedded into the nested `metadata` dict but the operators are unchanged:

```
>>> _default_filter_encoder({'a':1, '$or': [{'b':2}, {'c': 3}]})
{'metadata.a': 1, '$or': [{'metadata.b': 2}, {'metadata.c': 3}]}
```

hemidactylus: Let me add a note in the function, this is easily forgotten later on :)

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: Indeed the next phase of this initiative (a separate oncoming PR) will be an attempt to rationalize the flow for the various similarity_search* methods.

Made more complicated by the fact that most of the functions are public and (theoretically) entry points. But for now I would leave this structure as is, albeit suboptimal re: code duplication.

hemidactylus:
I want to draw your attention to the fact that this refactor is "imperfect", in that it leaks usage of `_id` partially in- and partially outside the classes.
Indeed, the Encoder classes ingest the doc id and are responsible for packaging it into the `_id` field of Astra document. They also return the `_id` within the projections they define.
But _id is also used directly in the vectorstore class.

Indeed an invariant for implementers should be to map the doc id to _id (descended from and Astra DB requirement).

Now, if you're happy with this, I'm fine with going on and perhaps refine this later (a slightly annoying task because it would mean kicking `_id` out of the class entirely ==> dict merging in the vectorstore for projections in a few places, subtleties with possible projections being `{"*": 1}` and so on).

libs/astradb/langchain_astradb/utils/encoders.py
```diff
@@ -0,0 +1,197 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+from langchain_core.documents import Document
+from typing_extensions import override
+
+
+def _default_filter_encoder(filter_dict: dict[str, Any]) -> dict[str, Any]:
```
cbornet, line 10: Rename it to a verb ?

libs/astradb/langchain_astradb/utils/encoders.py
```diff
@@ -0,0 +1,197 @@
...
+            content: textual content for the (LangChain) `Document`.
+            document_id: unique ID for the (LangChain) `Document`.
+            vector: a vector associated to the (LangChain) `Document`. This
+                parameter must be None for and only for server-side embeddings.
+            metadata: a metadata dictionary for the (LangChain) `Document`.
+
+        Returns:
+            a dictionary ready for storage onto Astra DB.
+        """
+        ...
```
cbornet, line 71: NIT: I think the `...` are not needed.

libs/astradb/langchain_astradb/utils/encoders.py
```diff
@@ -0,0 +1,197 @@
...
+    @override
+    def encode(
+        self,
+        content: str,
+        document_id: str,
+        vector: list[float] | None,
+        metadata: dict | None,
+    ) -> dict[str, Any]:
+        if vector is not None:
+            msg = "DefaultVectorize encoder cannot receive non-null vector"
```
cbornet, line 180: NIT: add value of vector in err message.

libs/astradb/langchain_astradb/utils/encoders.py
```diff
@@ -0,0 +1,197 @@
...
+    @override
+    def decode(self, astra_document: dict[str, Any]) -> Document:
+        return Document(
+            page_content=astra_document["content"],
+            metadata=astra_document["metadata"],
+        )
+
+    @override
+    def encode_filter(self, filter_dict: dict[str, Any]) -> dict[str, Any]:
+        return _default_filter_encoder(filter_dict)
```
cbornet, line 148: Could this be put in the parent class as a default method ? (and `_default_filter_encoder` method could then be removed)

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -891,18 +855,14 @@ async def aadd_texts(
...
+            embedding_vectors = [None for _ in list(texts)]
         else:
-            embedding_vectors = await self._get_safe_embedding().aembed_documents(
-                list(texts)
-            )
-            documents_to_insert = self._get_documents_to_insert(
-                texts, embedding_vectors, metadatas, ids
-            )
+            embedding_vectors = await self.embedding.aembed_documents(list(texts))  # type: ignore[union-attr]
```
cbornet, line 862: Can we fix the type issue instead of ignoring ?

cbornet: I think `_get_safe_embedding()` can be used for this ?

cbornet:
It's very nice to replace some of the if-else with OOP.
I wonder if we couldn't have `AbstractAstraDBVectorStore`, `AstraDBWithVectorizeVectorStore` and `AstraDBEmbeddingVectorStore` to simplify even more.

libs/astradb/langchain_astradb/utils/encoders.py
hemidactylus: I was a bit scared of doing that as it would suggest that this is ok "in most cases", while in practice the opposite is true (I am thinking of hypothetical crowds eager to define their own encoding here...).

Considering the "bring-your-own-collection" that additional implementations would refer to (likely flat in their fields), a better choice is perhaps to have a default that is the identity function and then set this one for both classes presently there ... which however would bring back the issue of sharing this function across both classes.

But given how easy is it to mistakenly use the wrong filter (and not see an actual error from find methods), I was leaning on not providing any default here.

But I'm open to re-evaluate of course!

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: My bad, an unwanted leftover. Fixed.

libs/astradb/langchain_astradb/utils/encoders.py
cbornet: Got it 👍

cbornet (approved):

PR #53, opened 2024-08-22
Activate ALL Ruff rules with some exclusions

cbornet:


libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1233,7 +1233,7 @@ def similarity_search(
         self,
         query: str,
         k: int = 4,
-        filter: dict[str, Any] | None = None,  # noqa: A002
+        filter: dict[str, Any] | None = None,
```
cbornet, line 1236: It is triggered with Ruff 0.5.0 but no longer with 0.5.7.
See https://github.com/astral-sh/ruff/pull/12415#issuecomment-2305300888

hemidactylus: TIL

hemidactylus (approved):

PR #61, opened 2024-08-27
Activate ruff rules BLE

cbornet:


libs/astradb/tests/integration_tests/test_vectorstores.py
```diff
@@ -74,7 +74,15 @@
 
 def is_nvidia_vector_service_available() -> bool:
     # For the time being, this is manually controlled
-    return os.getenv("NVIDIA_VECTORIZE_AVAILABLE") == "1"
+    if os.environ.get("NVIDIA_VECTORIZE_AVAILABLE"):
+        try:
+            # any non-zero counts as true:
+            return int(os.environ["NVIDIA_VECTORIZE_AVAILABLE"]) != 0
+        except (TypeError, ValueError):
```
cbornet, line 81: I don’t think it can be TypeError

cbornet: Then I would do ˋreturn int(os.getenv(NVIDIA, "0"))ˋ and remove the if clause.

hemidactylus (approved):

PR #63, opened 2024-08-28
Activate ruff rule C90

cbornet:


hemidactylus (approved):

libs/astradb/pyproject.toml
```diff
@@ -71,13 +71,12 @@ flake8-annotations.allow-star-arg-any = true
 flake8-annotations.mypy-init-return = true
 select = ["ALL"]
 ignore = [
-    "ANN", # Already checked by mypy
+    "C90", # Do we want to activate (complexity) ?
```
cbornet, line 74: The rule was added back.

PR #66, opened 2024-09-02
Some fixes to AstraDBVectorStore pydoc

cbornet:


hemidactylus:
Thank you for doing this!

PR #65, opened 2024-08-31
Vector store, autodetect mode

hemidactylus:
I will add some explanations these days (it's too late now, sorry).

PR #67, opened 2024-09-04
Implement GraphVectorStore for AstraDB

kerinin:
This adapts the CassandraGraphVectorStore to work with AstraDB. Fortunately, not schema changes are required to support the graph retrieval interface, so this should be a drop-in replacement for AstraDBVectorStore in cases where the graph retrieval functions are needed.


libs/astradb/langchain_astradb/_math.py
```diff
@@ -0,0 +1,48 @@
+"""Copied from langchain_community.utils.math.
+
+See https://github.com/langchain-ai/langchain/blob/langchain-community%3D%3D0.0.38/libs/community/langchain_community/utils/math.py .
+"""  # noqa: E501
+
+# ruff: noqa: EM102 TRY003
```
cbornet, line 6: Can these be fixed ?
`EM` rules can be fixed automatically with `poetry run ruff . --fix`

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,522 @@
+"""Astra DB graph vector store integration."""
+
+# ruff: noqa: FIX002 TD002 TD003 C901 D102 EM101 TRY003 SLF001 ANN003 UP007 ARG002
```
cbornet, line 3: Can these be fixed ?
Some rules can be auto-fixed with `poetry run ruff check . --fix --unsafe-fixes`.

libs/astradb/langchain_astradb/_math.py
```diff
@@ -0,0 +1,48 @@
...
+
+    x = np.array(x)
+    y = np.array(y)
+    if x.shape[1] != y.shape[1]:
+        raise ValueError(
+            f"Number of columns in X and Y must be the same. X has shape {x.shape} "
+            f"and Y has shape {y.shape}."
+        )
+    try:
+        import simsimd as simd
```
cbornet, line 32: Probably need to add simsimd to poetry's `typing` group.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,522 @@
...
+        # indexing policy setting
+        indexing_policy: dict[str, Any] = (
+            AstraDBVectorStore._normalize_metadata_indexing_policy(
+                metadata_indexing_include=metadata_indexing_include,
+                metadata_indexing_exclude=metadata_indexing_exclude,
+                collection_indexing_policy=collection_indexing_policy,
+            )
+        )
+
+        self.astra_env = _AstraDBCollectionEnvironment(
```
cbornet, line 116: could use `self.vectorstore.astra_env` ?

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,522 @@
...
+        async_astra_db_client: AsyncAstraDBClient | None = None,
+        namespace: str | None = None,
+        setup_mode: SetupMode = SetupMode.SYNC,
+        pre_delete_collection: bool = False,  # noqa: FBT001 FBT002
+        metric: str | None = None,
+        metadata_indexing_include: Iterable[str] | None = None,
+        metadata_indexing_exclude: Iterable[str] | None = None,
+        collection_indexing_policy: dict[str, Any] | None = None,
+        collection_vector_service_options: CollectionVectorServiceOptions | None = None,
+        collection_embedding_api_key: str | EmbeddingHeadersProvider | None = None,
```
cbornet, line 80: I think these are related to vectorize and vectorize can't be used with `AstraDBGraphVectorStore` ?

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,522 @@
...
+            collection_name=collection_name,
+            embedding=embedding,
+            metadata_indexing_include=metadata_indexing_include,
+            metadata_indexing_exclude=metadata_indexing_exclude,
+            collection_indexing_policy=collection_indexing_policy,
+        )
+
+        self.embedding_dimension: int | None = None
+        embedding_dimension: int | Awaitable[int] | None = None
+        if self.embedding is not None:
```
cbornet, line 101: Per typing, `self.embedding` cannot be None

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,522 @@
...
+            pre_delete_collection=pre_delete_collection,
+            embedding_dimension=embedding_dimension,
+            metric=metric,
+            requested_indexing_policy=indexing_policy,
+            default_indexing_policy=DEFAULT_INDEXING_OPTIONS,
+            collection_vector_service_options=collection_vector_service_options,
+            collection_embedding_api_key=collection_embedding_api_key,
+        )
+
+    def _get_safe_embedding(self) -> Embeddings:
```
cbornet, line 133: Per typing, `self.embedding` cannot be None. So we don't need `_get_safe_embedding` to please mypy.

cbornet:
It looks like we could just make `AstraDBVectorStore` extend GVS directly ?

libs/astradb/langchain_astradb/graph_vectorstores.py
cbornet: Maybe we can, I'm not sure. Is it possible to use vectorize to compute an embedding without storing it as a regular embedding service ?
We need an embedding service for the MMR traversal.
For AstraDBVectorStore, passing both embedding and collection_vector_service_options will raise an exception.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -126,20 +120,14 @@ def __init__(
             metric=metric,
             requested_indexing_policy=indexing_policy,
             default_indexing_policy=DEFAULT_INDEXING_OPTIONS,
-            collection_vector_service_options=collection_vector_service_options,
-            collection_embedding_api_key=collection_embedding_api_key,
+            collection_vector_service_options=None,
+            collection_embedding_api_key=None,
```
cbornet: NIT: those are the defaults so these lines can be removed

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -162,7 +150,7 @@ def embeddings(self) -> Embeddings | None:
     def add_nodes(
         self,
         nodes: Iterable[Node],
-        **kwargs,
+        **kwargs: Any, # noqa: ARG002
```
cbornet, line 83: Instead of the rule escape, it's possible to annotate the method with `@override` (`from typing_extensions import override`).
Ruff will see that the arg is there just to respect the interface contract and not enforce the rule.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -228,6 +218,7 @@ def similarity_search(
         metadata_filter: dict[str, Any] | None = None,
         **kwargs: Any,
     ) -> list[Document]:
+        """Return docs most similar to query."""
```
cbornet, line 151: The docstring is inherited from the parent class.
So instead of putting a docstring here, it would be better to annotate the method with `@override` so Ruff knows about it.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -237,6 +228,7 @@ def similarity_search_by_vector(
         metadata_filter: dict[str, Any] | None = None,
         **kwargs: Any,
     ) -> list[Document]:
+        """Return docs most similar to embedding vector."""
```
cbornet, line 161: Remove and annotate with `@override`

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -251,6 +243,7 @@ def traversal_search(
         metadata_filter: dict[str, Any] | None = None,
         **kwargs: Any,
     ) -> Iterable[Document]:
+        """Retrieve documents from traversing this graph store."""
```
cbornet, line 176: Remove and annotate with `@override`

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -339,6 +332,7 @@ def mmr_traversal_search(
         score_threshold: float = float("-inf"),
         metadata_filter: dict[str, Any] | None = None,
     ) -> Iterable[Document]:
+        """Retrieve documents from this graph store using MMR-traversal."""
```
cbornet, line 265: Remove and annotate with `@override`

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -355,7 +349,7 @@ def mmr_traversal_search(
         def get_adjacent(tags: set[str]) -> Iterable[_Edge]:
             targets: dict[str, _Edge] = {}
 
-            # TODO: Parralelize
```
cbornet: I don't have a strong opinion for TODOs.
Maybe for now, we could ignore `FIX002` globally and keep our TODOs.

libs/astradb/langchain_astradb/__init__.py
```diff
@@ -1,14 +1,17 @@
...
 from astrapy.info import CollectionVectorServiceOptions
 
 from langchain_astradb.cache import AstraDBCache, AstraDBSemanticCache
 from langchain_astradb.chat_message_histories import AstraDBChatMessageHistory
 from langchain_astradb.document_loaders import AstraDBLoader
-from langchain_astradb.graph_vectorstores import AstraDBGraphVectorStore
 from langchain_astradb.storage import AstraDBByteStore, AstraDBStore
 from langchain_astradb.vectorstores import AstraDBVectorStore
 
+from langchain_astradb.graph_vectorstores import AstraDBGraphVectorStore
```
cbornet, line 13: So the order of imports is important ?

libs/astradb/langchain_astradb/graph_vectorstores.py
cbornet: 👍 .
NIT: the rules could escape only the triggering line and not the full module.

libs/astradb/langchain_astradb/_math.py
```diff
@@ -0,0 +1,49 @@
+"""Copied from langchain_community.utils.math.
```
hemidactylus, line 1: `_math.py` and `utils/mmr.py` seem to share essentially the same "cosine_similarity" function. Probably the former can disappear and the latter used? (note mmr.py originated as a copy from langchain's main repo as well!)

libs/astradb/langchain_astradb/_mmr_helper.py
```diff
@@ -0,0 +1,255 @@
...
+        self.score = self.weighted_similarity - self.weighted_redundancy
+
+    def update_redundancy(self, new_weighted_redundancy: float) -> None:
+        if new_weighted_redundancy > self.weighted_redundancy:
+            self.weighted_redundancy = new_weighted_redundancy
+            self.score = self.weighted_similarity - self.weighted_redundancy
+
+
+class MmrHelper:
+    """Helper for executing an MMR traversal query.
```
hemidactylus, line 44: `_mmr_helper.py` as a whole seems to be specific to graph traversal (besides being mmr-related)? If so, suggestion is to rename the module so as to make it clear

libs/astradb/langchain_astradb/_math.py
```diff
@@ -0,0 +1,49 @@
...
+
+See https://github.com/langchain-ai/langchain/blob/langchain-community%3D%3D0.0.38/libs/community/langchain_community/utils/math.py .
```
hemidactylus, line 3: Generally the `_*.py` modules in langchain_astradb could probably be relocated in `utils/` (like similar modules are) ?

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,440 @@
...
+            query=query,
+            k=k,
+            filter=metadata_filter,
+            **kwargs,
+        )
+        visit_documents(0, docs)
+
+        return visited_docs
+
+    def mmr_traversal_search(
```
hemidactylus, line 252: It seems the lack of `**kwargs: Any` in this signature makes it incompatible with the method in parent `GraphVectorStore` class?

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,440 @@
...
+            return targets.values()
+
+        def fetch_neighborhood(neighborhood: Sequence[str]) -> None:
+            # Put the neighborhood into the outgoing tags, to avoid adding it
+            # to the candidate set in the future.
+            outgoing_tags.update({content_id: set() for content_id in neighborhood})
+
+            # Initialize the visited_tags with the set of outgoing from the
+            # neighborhood. This prevents re-visiting them.
+            visited_tags = self._get_outgoing_tags(neighborhood)
```
hemidactylus, line 330: I do not think I can see this method defined anywhere (a linter run on my side seems to confirm).

libs/astradb/langchain_astradb/__init__.py
hemidactylus: The graph module imports AstraDBVectorStore in turn. But luckily there is no "hard" circular import problem here. It is sufficient to change the import in `graph_vectorstores.py` from

```
from langchain_astradb import AstraDBVectorStore
```

to

```
from langchain_astradb.vectorstores import AstraDBVectorStore
```

so as to avoid referring to the partially-initialized `__init__` module when importing the AstraDBVectorStore. In my experimenting this solved the issue (without resorting to those pesky lazy imports in constructor).

libs/astradb/langchain_astradb/_mmr_helper.py
```diff
@@ -0,0 +1,255 @@
...
+    def _pop_candidate(self, candidate_id: str) -> NDArray[np.float32]:
+        """Pop the candidate with the given ID.
+
+        Returns:
+            The embedding of the candidate.
+        """
+        # Get the embedding for the id.
+        index = self.candidate_id_to_index.pop(candidate_id)
+        if not self.candidates[index].id == candidate_id:
+            raise ValueError(
```
hemidactylus, line 141: probably satisfying EM101 here would be a good thing (wouldn't it help with TRY003 as well?)

libs/astradb/langchain_astradb/_mmr_helper.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def _pop_candidate(self, candidate_id: str) -> NDArray[np.float32]:
+        """Pop the candidate with the given ID.
+
+        Returns:
+            The embedding of the candidate.
+        """
+        # Get the embedding for the id.
+        index = self.candidate_id_to_index.pop(candidate_id)
+        if not self.candidates[index].id == candidate_id:
```
hemidactylus, line 140: Suggestion: `if <val> != <test val>:`

libs/astradb/langchain_astradb/_mmr_helper.py
```diff
@@ -0,0 +1,255 @@
...
+    def __post_init__(self) -> None:
+        self.score = self.weighted_similarity - self.weighted_redundancy
+
+    def update_redundancy(self, new_weighted_redundancy: float) -> None:
+        if new_weighted_redundancy > self.weighted_redundancy:
+            self.weighted_redundancy = new_weighted_redundancy
+            self.score = self.weighted_similarity - self.weighted_redundancy
+
+
+class MmrHelper:
```
hemidactylus, line 43: So, if I get it right, this is a MMR reranking class but "dynamic", i.e. it supports adding entries a bunch at a time and it manages such late-time additions smartly. Very nice.
Incidentally, I wonder if eventually this class should be used in the MMR search methods of the `AstraDBVectorStore` itself (which now uses the much simpler, 25-line "static" algorithm in `utils/mmr.py`), just for the sake of avoiding logic duplication. _(filed for later consideration)._

libs/astradb/langchain_astradb/graph_vectorstores.py
hemidactylus: A note on supporting vectorize for graph vector stores: while Astra DB cannot be used as a string->vector service proper, there's a Data API flag (that astrapy understands) such that, each time an ANN query is sent (vectorize or otherwise), returns the "query vector used in the search" alongside the result set.

In other words, if the graph operations just need the vector for:

1. the original query sent to the store
2. the entries that are returned

then it should be possible to extend to vectorize.

For an idea on how to do that, see [these lines](https://github.com/langchain-ai/langchain-datastax/blob/2e664539817b11928bce101b2c3ee95eeca4964a/libs/astradb/langchain_astradb/vectorstores.py#L1546-L1555) (look for `query_vector`). Happy to discuss more!

libs/astradb/langchain_astradb/_mmr_helper.py
```diff
@@ -0,0 +1,255 @@
...
+    """
+    candidate_docs: dict[str, Document]
+    """Dict containing the documents associated with each candidate ID."""
+    candidate_embeddings: NDArray[np.float32]
+    """(N, dim) ndarray with a row for each candidate."""
+
+    best_score: float
+    best_id: str | None
+
+    def __init__(
```
hemidactylus, line 93: It would be nice to eventually have some unit tests for this class (it's quite some logic after all).

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,440 @@
+"""Astra DB graph vector store integration."""
+
+# ruff: noqa: SLF001 C901
```
hemidactylus, line 3: I would try to appease SLD001 (while I'd happily keep the complexity thing ignored for such a logic-dense module)

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,440 @@
...
+    def add_nodes(
+        self,
+        nodes: Iterable[Node],
+        **kwargs: Any, # noqa: ARG002
+    ) -> Iterable[str]:
+        """Add nodes to the graph store."""
+        docs = []
+        ids = []
+        for node in nodes:
+            node_id = secrets.token_hex(8) if not node.id else node.id
```
hemidactylus, line 89: Any specific reason to prefer this over `uuid.uuid4().hex` (which is what the AstraDBVectorStore does when no IDs are provided) ?
(to be clear, I have no preference - it is just a curiosity and a latent attraction to uniform solutions...)

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,440 @@
...
+        collection_indexing_policy: dict[str, Any] | None = None,
+    ):
+        """Create a new Graph Vector Store backed by AstraDB."""
+        self.link_to_metadata_key = link_to_metadata_key
+        self.link_from_metadata_key = link_from_metadata_key
+        self.content_id_key = content_id_key
+        self.session = None
+        self.embedding = embedding
+
+        self.vectorstore = AstraDBVectorStore(
```
hemidactylus, line 65: This could perhaps let additional kwargs pass through to the AstraDBVectorStore, for a "futureproof" usage (e.g. Environment for HCD, or "autodetect_collection" to facilitate heterogeneous LangFlow constructions etc)

libs/astradb/langchain_astradb/graph_vectorstores.py
kerinin: This sounds like it will work for this use case, however I think we should defer implementing it in this PR.

libs/astradb/langchain_astradb/__init__.py
kerinin: Exactly.

libs/astradb/langchain_astradb/_math.py
kerinin: I'll just use the existing implementation. This is copied from the previous implementation. We may want to pull in the `simsimd` work, but I'd treat that as a separate PR.

libs/astradb/langchain_astradb/_mmr_helper.py
kerinin: 🤷 I'm just copying it from a previous implementation 😄 . But yeah, I think that's right.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,478 @@
...
+
+        def get_adjacent(tags: set[str]) -> Iterable[_Edge]:
+            targets: dict[str, _Edge] = {}
+
+            # TODO: Would be better parralelized
+            for tag in tags:
+                metadata_parameter = self._filter_to_metadata(
+                    metadata_filter
+                )
+                metadata_parameter[f"metadata.{self.link_from_metadata_key}"] = tag
```
bjchambers, line 323: It's not clear this will work. AFAICT, this is doing an equality query on link from metadata key. It needs to be a "contains" query. Specifically, the "link_from_metadata_key" may be a set of tags, and we want to find any node that has `tag` in the corresponding set.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,478 @@
...
+
+        # For each unselected node, stores the outgoing tags.
+        outgoing_tags: dict[str, set[str]] = {}
+
+        visited_tags: set[str] = set()
+
+        def get_adjacent(tags: set[str]) -> Iterable[_Edge]:
+            targets: dict[str, _Edge] = {}
+
+            # TODO: Would be better parralelized
```
bjchambers, line 318: Will be significantly slower without parallelization...

kerinin: Agreed - hence the TODO :)

libs/astradb/langchain_astradb/graph_vectorstores.py
hemidactylus: I can confirm that the same filter will match both documents with the required tag being one in the list (as intended), and also documents with just that tag instead of a list (which should never exist if the collection is in the sole hands of this store):

```
>>> for doc in t.find({"metadata.tags": "my_tag"}, projection={"_id": False}):
...   print(doc)
... 
{'metadata': {'tags': 'my_tag'}}
{'metadata': {'tags': ['one_tag', 'my_tag', 'another_tags']}}
```

So this is good as is IMO.

libs/astradb/langchain_astradb/_math.py
hemidactylus: @kerinin, I'm coming back to this point for a crucial issue. I can confirm the current code does not use `simsimd` properly and breaks with a TypeError as is now.
(note: as this affects a module not involved in the PR, github does not let me comment on it).

In my test runs, it appears that, to avoid a `TypeError`, line 53 of `mmr.py` should change from

```
        z = 1 - simd.cdist(x, y, metric="cosine")
```

to 

```
        z = 1 - np.array(simd.cdist(x, y, metric="cosine"))
```

Please check.

The reason this was never caught is that until now the `simsimd` package had no reason to be installed so the `cosine_similarity` function went into the "just numpy" branch of the two.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,478 @@
...
+        store = cls(embedding, **kwargs)
+        store.add_documents(documents, ids=ids)
+        return store
+
+    @override
+    def similarity_search(
+        self,
+        query: str,
+        k: int = 4,
+        metadata_filter: dict[str, Any] | None = None,
```
hemidactylus, line 153: Is it intentional that (for the better or the worse) this parameter, following LC's abstract interface, is called `filter` in the vector store (while in this class it is called `metadata_filter` consistently across all methods?)

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,478 @@
...
+
+
+class AstraDBGraphVectorStore(GraphVectorStore):
+    def __init__(
+        self,
+        embedding: Embeddings,
+        collection_name: str,
+        link_to_metadata_key: str = "links_to",
+        link_from_metadata_key: str = "links_from",
+        content_id_key: str = "content_id",
```
hemidactylus, line 54: From a cursory inspection of the module, it would seem that this concept can disappear in favour of using the documents' `_id` directly. Would that clash with the contract with the abstract GraphVectorStore? (it could make things easier re: duplication of info otherwise...)

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,478 @@
...
+
+                    # If we can continue traversing from this document,
+                    if d < depth and doc.metadata[self.link_to_metadata_key]:
+                        # Record any new (or newly discovered at a lower depth)
+                        # tags to the set to traverse.
+                        for tag in doc.metadata[self.link_to_metadata_key]:
+                            if d <= visited_tags.get(tag, depth):
+                                # Record that we'll query this tag at the
+                                # given depth, so we don't fetch it again
+                                # (unless we find it an earlier depth)
```
hemidactylus, line 218: "at" ?

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,478 @@
...
+        outgoing_tags: dict[str, set[str]] = {}
+
+        visited_tags: set[str] = set()
+
+        def get_adjacent(tags: set[str]) -> Iterable[_Edge]:
+            targets: dict[str, _Edge] = {}
+
+            # TODO: Would be better parralelized
+            for tag in tags:
+                metadata_parameter = self._filter_to_metadata(
```
hemidactylus, line 320: It cannot be excluded that (some future implementation of) _filter_to_metadata will return a _reference_ to its argument, unchanged. Hence I strongly suggest to do a copy here to avoid future sneaky bugs from unwittingly mutating the original `metadata_filter` argument  to mmr_traversal_search.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,478 @@
...
+                    metadata_filter
+                )
+                metadata_parameter[f"metadata.{self.link_from_metadata_key}"] = tag
+
+
+                hits = list(self.astra_env.collection.find(
+                    filter=metadata_parameter,
+                    projection={
```
hemidactylus, line 328: For custom-scheme documents this would break. Suggest to always pass through the "encoder" (renamed "codecs" in the latest on main meanwhile).
Right now this would have to become:

```
projection=self.vectorstore.document_encoder.base_projection
```

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,478 @@
...
+                        adjacent.target_text_embedding,
+                    )
+            helper.add_candidates(new_candidates)
+
+        def fetch_initial_candidates() -> None:
+            metadata_parameter = self._filter_to_metadata(metadata_filter)
+            hits = list(
+                self.astra_env.collection.find(
+                    filter=metadata_parameter,
+                    projection={
```
hemidactylus, line 391: For custom-scheme documents this would break. Suggest to always pass through the "encoder" (renamed "codecs" in the latest on main meanwhile).
Right now this would have to become:

```
projection=self.vectorstore.document_encoder.base_projection
```

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,478 @@
...
+
+        Args:
+            source_ids: The IDs of the source nodes to retrieve outgoing tags for.
+        """
+        tags = set()
+
+        for source_id in source_ids:
+                hits = list(self.astra_env.collection.find(
+                    filter={f"metadata.{self.content_id_key}": source_id},
+                    projection={
```
hemidactylus, line 277: This assumes that metadata is a subdict on the document - an assumption that becomes problematic with the "autodetect" mode that was just added to the underlying vector store.

Here (and possibly in a few other places) there is a risk in directly working on the documents rather than going through the "encoder" layer (recently renamed to "codecs" after #65 got merged).

(related: see remark on lines 328, 391; but e.g. also line 276 here above might become problematic for that same assumption).

My suggestion is to merge the latest main and expand the scope of the codecs as necessary to always pass through that layer, hence gaining resilience against arbitrary-schema v.stores. We can discuss the details, it;s not that scary :)

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -0,0 +1,478 @@
...
+                    content_id = hit["metadata"][self.content_id_key]
+
+                    if content_id not in targets:
+                        targets[content_id] = _Edge(
+                            target_content_id=content_id,
+                            target_text_embedding=vector,
+                            target_link_to_tags=set(
+                                hit.get(self.link_to_metadata_key, [])
+                            ),
+                            target_doc=Document(
```
hemidactylus, line 351: In the same spirit as "defer to the codec", this would have to be something like:

```
target_doc=self.vectorstore.document_encoder.decode(hit)
```

libs/astradb/langchain_astradb/graph_vectorstores.py
cbornet: There is no `filter` param in the parent VectorStore interface. It only appears in child implementations with various formats that depend on the underlying tech.
I think `metadata_filter` was used to avoid collision with the Python builtin `filter()` function.

kerinin:
Going to go ahead and merge with the integration test failure - I don't think it's related to these changes.

libs/astradb/langchain_astradb/graph_vectorstores.py
hemidactylus: Sorry, my bad then. All good.

PR #68, opened 2024-09-05
astradb[patch]: prep for langchain-core 0.3

baskaryan:
remove any pydantic v1 use in preparation for langchain-core 0.3, which uses pydantic v2


hemidactylus (approved):
LGTM, thank you for this!

PR #75, opened 2024-09-18
Add AstraDBGraphVectorStore testing

hemidactylus:
Adding integration+unit test for AstraDBGraphVectorStore.

Doing so uncovered unexpected behaviour w.r.t the "id in document vs. ids passed as standalone" in the from_documents implementation for both AstraDBGraphVectorStore and AstraDBVectorStore, which this PR also fixes (and tests).

Incidentally the "filter to metadata" utility method was made public to allow the Graph layer to defer to its VectorStore delegate instead of reimplementing it.

This PR also addresses a [comment](https://github.com/langchain-ai/langchain-datastax/pull/67#discussion_r1759606376) to #67 that was unattended, inadvertently bringing a critical bug into main.


libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -139,7 +139,17 @@ def from_documents(
...
         """Return GraphVectorStore initialized from documents and embeddings."""
         store = cls(embedding, **kwargs)
-        store.add_documents(documents, ids=ids)
+        # `store.add_documents` ends up calling store.add_nodes, which
+        # discards the kwargs including ids. This is the place to normalize
+        # the documents' .id and the separate ids into one and the same:
+        _documents: Iterable[Document]
+        if ids is not None:
+            _documents = [document.copy() for document in documents]
+            for _doc_id, _document in zip(ids, _documents):
```
cbornet, line 148: There should be some check that `ids` and `documents` have the same size.

hemidactylus: Closing since `ids` is going away from this method.

_Side note for your amusement_. This check would be not entirely trivial if one does not want to `list(...)` the iterables. I had started to implement something like [this](https://github.com/langchain-ai/langchain/blob/4e0a6ebe7d6a3f404077817e483a1fb56b021d88/libs/core/langchain_core/graph_vectorstores/base.py#L76-L100) before removing it all.

But the real caveat is that `zip(ite1, ite2)` treats the two iterables differently (i.e. it's implemented straigthforwardly). It can consume one item too much from ite1 which goes nowhere. The first snippet here prints "4", the second one gives a StopIteration:

```
r1 = (i for i in range(4))
r2 = (i for i in range(5))

for p in zip(r1, r2):
   print(p)


next(r2)
```

```
r1 = (i for i in range(4))
r2 = (i for i in range(5))

for p in zip(r2, r1):
   print(p)


next(r2)
```

cbornet (approved):
LGTM!

hemidactylus:
CI is having lots of troubles (hitting DB limits and whatnot; I'll have a PR to simplify, rationalize and optimize the tests and the way they use DB resources).
Meanwhile, I ran all tests locally, piecewise to avoid overloading the DB, and it all passes.
Merging now.

PR #76, opened 2024-09-20
Better handling of errors during insert many (Vector Store)

hemidactylus:
The add_texts logic is two-steps:

1. try insert many (the Data API does not allow overwrites)
2. collect failed insertions and fanout concurrent one-by-one "updates" for preexisting ids.

Until now, any failing insertion in step 2 was taken to mean "preexisting" which of course is not necessarily the case. This PR fixes this, failing fast if other error code are obtained from the Data API in step 1. This also improves the clarity of possible errors that can occur in step 2 since it removes silly situations where the update is doomed to fail.

### DIFF
--- a/libs/astradb/langchain_astradb/vectorstores.py
+++ b/libs/astradb/langchain_astradb/vectorstores.py
@@ -63,6 +63,8 @@ DocDict = Dict[str, Any]  # dicts expressing entries to insert
 
 # indexing options when creating a collection
 DEFAULT_INDEXING_OPTIONS = {"allow": ["metadata"]}
+# error code to check for during bulk insertions
+DOCUMENT_ALREADY_EXISTS_API_ERROR_CODE = "DOCUMENT_ALREADY_EXISTS"
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1107,13 +1117,21 @@ async def aadd_texts(
...
-            inserted_ids = err.partial_result.inserted_ids
-            inserted_ids_set = set(inserted_ids)
-            ids_to_replace = [
-                document["_id"]
-                for document in documents_to_insert
-                if document["_id"] not in inserted_ids_set
-            ]
+            # check that the error is solely due to already-existing documents
+            error_codes = {
+                getattr(err_desc, "error_code", None)
```
cbornet, line 1122: Why not use `err_desc.errorcode` directly ?

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1107,13 +1117,21 @@ async def aadd_texts(
...
+                for err_desc in err.error_descriptors
+            }
+            if error_codes == {DOCUMENT_ALREADY_EXISTS_API_ERROR_CODE}:
```
cbornet, line 1125: Is it on purpose to raise if there are several error codes including the document already exists one ?

cbornet (approved):

PR #80, opened 2024-09-25
Include more information about the caller

kerinin:
Currently, when making calls to AstraDB, we include a `User-Agent` header identifying the API caller. This header is currently set to "langchain" by default at the moment. This change threads the `caller_name` parameter throughout the API and appends a suffix identifying which class created the API client:

* langchain/cache
* langchain/chat_message_history
* langchain/document_loader
* langchain/graph_vectorstore
* langchain/semantic_cache
* langchain/vectorstore

This metadata will be useful for debugging and tracking utilization of different library features.

NOTE: This change will change the value of User-Agent headers, for
example from `langchain/0.3.6` to `langchain/vectorstore/0.3.6`


hemidactylus:
This should be discussed a bit more before merging, as it involves possible changes across several layers (e.g. usage analytics automation, etc). Let us first ensure everyone is aligned on how we want to achieve this.

hemidactylus:
Closed. Superseded by #96

PR #84, opened 2024-09-30
Remove some  rules escapes

cbornet:


libs/astradb/tests/integration_tests/test_vectorstore_ddl_tests.py
```diff
@@ -489,15 +486,14 @@ async def test_astradb_vectorstore_indexing_custom_async(
     def test_astradb_vectorstore_vectorize_headers_precedence_stringheader(
         self,
         astra_db_credentials: AstraDBCredentials,
-        collection_vz: Collection,  # noqa: ARG002
```
cbornet, line 492: Injecting `collection_vz` had no use here ?

hemidactylus: Ah, I think you spotted a bug in testing (both here and in the subsequent `test_astradb_vectorstore_vectorize_headers_precedence_headerprovider`). Well done!

hemidactylus: Pushed a fix.

hemidactylus (approved):
LGTM, thanks for cleaning the noqas and also replacing the constants with `<collection>.name` !

PR #88, opened 2024-10-01
Automatic environment from data api endpoint (if not supplied)

hemidactylus:
This PR enables automatic detection of the "environment" for using the Data API, inferred from the API Endpoint when the specific `environment` parameter is left unspecified.

**This makes sense here, since - as opposed to a broader astrapy usage - no actual "admin" operations are performed, so there's no harm in having the endpoint first and creating the instances later.**

This holds throughout all LC objects since the new logic takes place in the _`AstraDBEnvironment`.

Added a test for this logic.


mfortman11 (approved):
Super helpful! TY

PR #89, opened 2024-10-01
Add [a]delete_by_metadata and [a]update_metadata methods to vector store

hemidactylus:
This PR adds four methods to the vector store, the a/sync versions of:

- update metadata (of several documents at once). This has the semantic of updating a dictionary, i.e. old keys are not lost and overlapping keys are overridden;
- delete (a number of documents) by a metadata filter.

These methods are needed by the operation of "promoting" a vector store to a graph vector store.

### DIFF
--- a/libs/astradb/langchain_astradb/vectorstores.py
+++ b/libs/astradb/langchain_astradb/vectorstores.py
@@ -1222,6 +1222,110 @@ class AstraDBVectorStore(VectorStore):
                 raise ValueError(msg)
         return inserted_ids
 
+    def update_metadata(
+        self,
+        id_to_metadata: dict[str, dict],
+        *,
+        overwrite_concurrency: int | None = None,
+    ) -> int:
+        """Add/overwrite the metadata of existing documents.
+
+        For each document to update, the new metadata dictionary is added
+        to the existing metadata, overwriting individual keys that existed already.
+
+        Args:
+            id_to_metadata: map from the Document IDs to modify to the
+                new metadata for updating. Keys in this dictionary that
+                do not correspond to an existing document will be silently ignored.
+                The values of this map are metadata dictionaries for updating
+                the documents. Any pre-existing metadata will be merged with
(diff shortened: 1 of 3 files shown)


hemidactylus:
@epinzur FYI

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -827,6 +827,62 @@ async def adelete(
...
+            implying emptying the store entirely.
+        """
+        self.astra_env.ensure_db_setup()
+        metadata_parameter = self.filter_to_query(filter)
+        del_result = self.astra_env.collection.delete_many(
+            filter=metadata_parameter,
+        )
+        if del_result.deleted_count is not None and del_result.deleted_count >= 0:
+            return del_result.deleted_count
+        return None
```
epinzur, line 856: could we throw an error if `len(filter) == 0` instead of allowing it to delete everything?

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1166,6 +1222,110 @@ async def _replace_document(
...
 
+    def update_metadata(
+        self,
+        id_to_metadata: dict[str, dict],
+        *,
+        overwrite_concurrency: int | None = None,
+    ) -> int:
+        """Add/overwrite the metadata of existing documents.
+
+        For each document to update, the new metadata dictionary is added
```
epinzur, line 1233: maybe `appended` instead of `added`? I think it would be slightly clearer.

epinzur (approved):

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: Fine for me (the reason for allowing `{}` is that we were asked to remove a similar guardrail in astrapy's delete_many. But I appreciate that here the scope is much more limited and safer is presumably better).

I'll change it shortly.

PR #92, opened 2024-10-04
WIP: upgrade standard RAG to Graph RAG

epinzur:


libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -87,31 +106,41 @@ def add_nodes(
...
-                metadata=metadata,
+                metadata=node.metadata,
                 id=node_id,
             )
             docs.append(doc)
             ids.append(node_id)
 
         return self.vectorstore.add_documents(docs, ids=ids)
 
+    def upgrade_chunks(
```
bjchambers, line 121: This should probably be `upgrade_documents` or something to match naming. I'm also not sure about this API that iterates over all the documents. We may be better off with something with building blocks:

1. Iterate over documents (with an optional filter)
2. Function for adding links to already persisted documents. Something like `add_links({ "id1": [new_link1, new_link2], ... })`.

This would allow the user to control which documents to add links to. It would also be useful after the initial upgrade, since the `add_links` method could be used for adding links discovered later, etc.

epinzur:
will replace this with a different PR

PR #95, opened 2024-10-08
Update AstraDBGraphVectorStore to match implementation of CassandraGraphVectorStore

epinzur:
This change corresponds with https://github.com/langchain-ai/langchain/pull/27253 which implements changes for `CassandraGraphVectorStore`.

After both are merged, both implementations will be nearly identical, and we can look into creating some shared code.


libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -22,6 +22,7 @@
 
 import numpy as np
 from astrapy.exceptions import InsertManyException
+from langchain_community.vectorstores.utils import maximal_marginal_relevance
```
hemidactylus, line 25: there are (very tiny) differences between the definition in the module that this PR is deleting and the one imported from community. I have no idea if these are relevant (perhaps in some corner cases, dim=1 vectors or something...?)

libs/astradb/tests/unit_tests/test_mmr_helper.py
```diff
@@ -46,22 +48,96 @@ def test_mmr_helper_functional(self) -> None:
...
+        # Fetching the 2 nearest neighbors to 0.0
+        helper.add_candidates(
+            {
+                "v0": self.angular_embedding(-0.124),
+                "v1": self.angular_embedding(+0.127),
+            }
+        )
+        assert helper.pop_best() == "v0"
+
+        # After v0 is seletected, new nodes are discovered.
```
hemidactylus, line 126: "seletected" :)

libs/astradb/tests/unit_tests/test_mmr_helper.py
```diff
@@ -46,22 +48,96 @@ def test_mmr_helper_functional(self) -> None:
...
+        """Test end to end construction and MMR search.
+        The embedding function used here ensures `texts` become
+        the following vectors on a circle (numbered v0 through v3):
+
+            ______ v2
+            /      \
+            /        |  v1
+        v3  |     .    | query
+            |        /  v0
+            |______/                 (N.B. very crude drawing)
```
hemidactylus, line 106: this diagram looks like it lost some spacing becoming harder to read

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -32,28 +36,76 @@
...
+            # Let the base class default method raise the TypeError
+            return super().default(obj)
+
+    return json.dumps(links, cls=SetAndLinkEncoder)
+
+
+def _deserialize_links(json_blob: str | None) -> set[Link]:
+    return {
+        Link(kind=link["kind"], direction=link["direction"], tag=link["tag"])
+        for link in cast(list[dict[str, Any]], json.loads(json_blob or ""))
```
hemidactylus, line 72: Is it intended that `json.loads(json_blob or "")`  breaks in the "or" case? Perhaps

```
json_blob or "[]"
```

is a safer fallback (= no links in case blob is empty)?

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -161,11 +212,24 @@ def __init__(
...
         """
-        self.link_to_metadata_key = link_to_metadata_key
-        self.link_from_metadata_key = link_from_metadata_key
+        self.metadata_incoming_links_key = metadata_incoming_links_key
         self.embedding = embedding
 
-        self.vectorstore = AstraDBVectorStore(
+        # update indexing policy to ensure incoming_links are indexed, and the
+        # full links blob is not.
+        if collection_indexing_policy is not None:
```
hemidactylus, line 220: The goal of this portion is to ensure that graphization works indexed even on a pre-existing store?
If so, this wouldn't unfortunately be effective (indexing is specified at collection _creation_ time, cannot be changed later in Astra DB. Changing the index and retrying a createCollection will give a server error if mismatching settings).

For preexisting collections: one has to hope that the pre-graph user created the collection with a future-compatible indexing (i.e. the default one, or anyway a blacklisting one rather than whitelisting), or the graphization will work on nonindexed paths.

When this path is creating the collection anew, aren't the default settings from the vector store (i.e. index all of metadata) good enough already?

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -189,48 +253,74 @@ def __init__(
             autodetect_collection=autodetect_collection,
         )
 
-        self.astra_env = self.vectorstore.astra_env
+        self.astra_env = self.vector_store.astra_env
 
     @property
     @override
-    def embeddings(self) -> Embeddings | None:
+    def embeddings(self) -> Embeddings:
```
hemidactylus, line 260: Parent (LC core's `VectorStore`) has

```
def embeddings(self) -> Optional[Embeddings]:
```

which may come handy if left as is for a future $vectorize extension of this class.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -161,11 +212,24 @@ def __init__(
                 ``metric``, ``setup_mode``, ``metadata_indexing_include``,
                 ``metadata_indexing_exclude``, ``collection_indexing_policy``.
         """
-        self.link_to_metadata_key = link_to_metadata_key
-        self.link_from_metadata_key = link_from_metadata_key
+        self.metadata_incoming_links_key = metadata_incoming_links_key
```
hemidactylus, line 215: do I get this right? Since the traversal is only following the arrows in reverse, there's no need to store the "outgoing" as indexed metadata, rather it's enough to lump them in the "links" inert metadata portion. Correct?

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -189,48 +253,74 @@ def __init__(
...
+
+        Returns:
+            The same Document with restored links.
+        """
+        links = _deserialize_links(doc.metadata.get(METADATA_LINKS_KEY))
+        doc.metadata[METADATA_LINKS_KEY] = links
+        del doc.metadata[self.metadata_incoming_links_key]
+        return doc
+
+    # TODO: Async (aadd_nodes)
```
hemidactylus, line 291: is this TODO still todo?

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -595,7 +689,355 @@ def fetch_initial_candidates() -> None:
...
+        for result in results:
+            for doc, embedding, doc_id in result:
+                if doc_id not in retrieved_docs:
+                    retrieved_docs[doc_id] = doc
+                if doc_id not in targets:
+                    node = _doc_to_node(doc=doc)
+                    targets[doc_id] = AdjacentNode(node=node, embedding=embedding)
+
+        # TODO: Consider a combined limit based on the similarity and/or
+        # predicated MMR score?
```
hemidactylus, line 1042: "predicted" perhaps? (or, perhaps, I did not get what that means)

hemidactylus:
You may find it convenient to merge the latest main, which introduces management for a new scheme of "callers" (i.e. user-agent management down to the astrapy layer). For the gvs.py file, this amounts to accepting (and passing down to the VectorStore) a couple of kwargs more.

libs/astradb/langchain_astradb/graph_vectorstores.py
epinzur: good catch

libs/astradb/langchain_astradb/graph_vectorstores.py
epinzur: I was trying to avoid indexing the links metadata if possible. But I'm fine removing this block.

libs/astradb/langchain_astradb/graph_vectorstores.py
epinzur: will revert

libs/astradb/langchain_astradb/graph_vectorstores.py
epinzur: yes, external to this PR.  Same on the CassandraGraphVectorStore side.

libs/astradb/langchain_astradb/graph_vectorstores.py
epinzur: simplified this and added new a integration test: `test_upgrade_to_graphvectorstore.py`

libs/astradb/tests/integration_tests/test_upgrade_to_graphvectorstore.py
```diff
@@ -0,0 +1,146 @@
...
+        ],
+        ids=[
+            "default_store_upgrade_should_succeed",
+            "allow_store_upgrade_with_allow_policy_should_fail",
+            "allow_store_upgrade_with_no_policy_should_fail",
+            "deny_store_upgrade_with_deny_policy_should_succeed",
+            "deny_store_upgrade_with_no_policy_should_fail",
+        ],
+    )
+    def test_upgrade_to_gvs(
```
epinzur, line 102: here is an initial tests that tries upgrading a vector store to a graph vector store, with various indexing sessions.

I think this covers enough cases for now.  I can add more when I add the update documents code in the next PR.

libs/astradb/langchain_astradb/utils/astradb.py
```diff
@@ -380,18 +380,20 @@ def __init__(
...
+                try:
+                    if not self._validate_indexing_policy(
+                        collection_descriptors=collection_descriptors,
+                        collection_name=self.collection_name,
+                        requested_indexing_policy=requested_indexing_policy,
+                        default_indexing_policy=default_indexing_policy,
+                    ):
+                        raise data_api_exception  # noqa: TRY201
+                except ValueError as validation_error:
+                    raise validation_error from data_api_exception
```
epinzur, line 396: this was throwing a 2nd exception from inside the exception handling. seems better to wrap the first exception.

libs/astradb/langchain_astradb/graph_vectorstores.py
hemidactylus: ¯\_(ツ)_/¯

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -167,79 +246,174 @@ def __init__(
...
+                async_astra_db_client=async_astra_db_client,
+            )
+
+            # # attempt a query to see if the table is setup correctly
 
-        self.astra_env = self.vectorstore.astra_env
+            # self.metadata_search(filter = {
+            #     self.metadata_incoming_links_key : "test"
+            # }, n=1)
+        except BaseException as exp:
```
hemidactylus, line 315: I _think_ you can catch the narrower `astrapy.exception.DataAPIException` here since the only case this block is after is the unindexed-path problem.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -167,79 +246,174 @@ def __init__(
...
+                ext_callers=ext_callers,
+                component_name=component_name,
+                astra_db_client=astra_db_client,
+                async_astra_db_client=async_astra_db_client,
+            )
+
+            # # attempt a query to see if the table is setup correctly
 
-        self.astra_env = self.vectorstore.astra_env
+            # self.metadata_search(filter = {
```
hemidactylus, line 312: Is this part commented on purpose?

- if so, can remove
- but I think a search should be executed here to catch the unindexed-path problems. One may have used setupMode=OFF (no collection creation, no indexing-mismatch errors detected otherwise).

Now a bold idea: perhaps one can get away with fewer complications around "checking the indexing settings pre-instantiations" and just run this puny search. This would be an 'experimental' check that the desired path is indexed indeed, post-facto, guaranteed to be the truth.

A caveat is that probably there should be an if setup_mode==SetupMode.ASYNC to run either an `ametadata_search` or (if OFF or SYNC) a regular `metadata_search`.

disclaimer: I did not verify this idea.

libs/astradb/tests/integration_tests/test_upgrade_to_graphvectorstore.py
```diff
@@ -0,0 +1,146 @@
...
+        request: pytest.FixtureRequest,
+    ) -> None:
+        # Create Vector Store, load a document
+        v_store: AstraDBVectorStore = request.getfixturevalue(store_name)
+        doc_id = "AL"
+        doc_al = Document(id=doc_id, page_content="[-1, 9]", metadata={"label": "AL"})
+        v_store.add_documents([doc_al])
+
+        # Try to create a GRAPH Vector Store using the existing collection from above
+        try:
```
hemidactylus, line 120: Suggestion: refactor this test to take advantage of the `pytest.raises` test (shorter, better match logic, etc).
See e.g. [here](https://github.com/langchain-ai/langchain-datastax/blob/31c25d4263b67ec47d66d64e88417b0dd9da145e/libs/astradb/tests/integration_tests/test_document_loaders.py#L345).

libs/astradb/tests/integration_tests/test_upgrade_to_graphvectorstore.py
hemidactylus: After spending days fighting test flakiness a couple of weeks ago, I can offer a suggestion: using the same collection name and repeatedly creating-and-deleting it with different settings (incl. indexing) is likely going be a strong source of flakiness.

My suggestion is as follows: while still deleting each collection at the end of each test case for this function, better use different collection names, one per each of the indexing configured at create_collection time.

If I see this correctly, that would be three: (1) vectorstore default, (2) a denylist, (3) an allowlist. Now, you can re-use existing ephemeral collections for (1) and (1) - these would be `EPHEMERAL_DEFAULT_IDX_NAME_D2` and `EPHEMERAL_CUSTOM_IDX_NAME_D2`. You may want to add a new, deny-list-based ephemeral analogous collection name here for (3), create its fixtures, and add it to the [ephemeral_indexing_collections_cleaner](https://github.com/langchain-ai/langchain-datastax/blob/31c25d4263b67ec47d66d64e88417b0dd9da145e/libs/astradb/tests/integration_tests/conftest.py#L509) "cleaner" fixture.

The gist is: if each time a collection with a given name is created with the same settings (incl. indexing), you won't run into those pesky metadata caching issues, source of hard-to-track (and undeserved) bugs!
(note: this implies that the collection with custom indexing is created exactly [like it is now](https://github.com/langchain-ai/langchain-datastax/blob/31c25d4263b67ec47d66d64e88417b0dd9da145e/libs/astradb/tests/integration_tests/test_vectorstore_ddl_tests.py#L296), ... but looking at what this new test does it does not seem to be an issue).

libs/astradb/langchain_astradb/graph_vectorstores.py
epinzur: I don't think we can get away with just checking this search since the collection initialization code throws an exception in the SYNC case.

I'll add test cases for both ASYNC and OFF, which you correctly note will need handling around the search.

libs/astradb/tests/integration_tests/test_upgrade_to_graphvectorstore.py
epinzur: split into `_success` and `_failure` tests based on my expectation if it should succeed or fail, to make it easier to user `pytest.raises`

libs/astradb/tests/integration_tests/test_upgrade_to_graphvectorstore.py
epinzur: done

libs/astradb/langchain_astradb/graph_vectorstores.py
epinzur: changed to catch `ValueError` as I saw both that and `DataAPIException` in testing.  And `DataAPIException` inherits from `ValueError`.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -167,79 +246,174 @@ def __init__(
...
+            #     self.metadata_incoming_links_key : "test"
+            # }, n=1)
+        except BaseException as exp:
+            # determine if error is because of a un-indexed column. Ref:
+            # https://docs.datastax.com/en/astra-db-serverless/api-reference/collections.html#considerations-for-selective-indexing
+            error_message = str(exp).lower()
+            if ("unindexed filter path" in error_message) or (
+                "incompatible with the requested indexing policy" in error_message
+            ):
+                msg = "The collection configuration is incompatible with vector graph store. Please create a new collection."  # noqa: E501
```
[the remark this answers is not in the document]
epinzur: updated

libs/astradb/tests/unit_tests/test_callers.py
```diff
@@ -294,6 +294,58 @@ def test_callers_component_vectorstore(self, httpserver: HTTPServer) -> None:
...
+            {
+                "status": {
+                    "collections": [
+                        {"name": "my_graph_coll", "options": {"vector": {"dimension": 2}}}
+                    ]
+                },
+                "data": {
+                    "nextPageState": None,
+                    "documents": [],
+                }
```
epinzur, line 334: is this correct?  I built it up manually until the test passed.

libs/astradb/tests/unit_tests/test_callers.py
```diff
@@ -294,6 +294,63 @@ def test_callers_component_vectorstore(self, httpserver: HTTPServer) -> None:
...
+                            "options": {"vector": {"dimension": 2}},
+                        }
+                    ]
+                },
+                "data": {
+                    "nextPageState": None,
+                    "documents": [],
+                },
+            }
+        )
```
epinzur, line 343: is this okay?  I built it up manually until the test passed.

libs/astradb/tests/conftest.py
```diff
@@ -33,7 +33,6 @@ def embed_query(self, text: str) -> list[float]:
         try:
             vals = json.loads(text)
         except json.JSONDecodeError:
-            print(f'[ParserEmbeddings] Returning a moot vector for "{text}"')
```
epinzur, line 36: okay to drop this?

libs/astradb/tests/integration_tests/test_graphvectorstore.py
```diff
@@ -226,7 +225,35 @@ def test_gvs_similarity_search(
         ],
         ids=["native_store", "autodetected_store"],
     )
-    def test_gvs_traversal_search(
+    async def test_gvs_similarity_search_async(
```
epinzur, line 228: added async versions of everything to tests

libs/astradb/tests/integration_tests/test_graphvectorstore.py
```diff
@@ -272,7 +325,159 @@ def test_gvs_mmr_traversal_search(
...
+
+    @pytest.mark.parametrize(
+        ("store_name"),
+        [
+            ("populated_graph_vector_store_d2"),
+            ("autodetect_populated_graph_vector_store_d2"),
+        ],
+        ids=["native_store", "autodetected_store"],
+    )
+    def test_gvs_metadata_search_sync(
```
epinzur, line 370: added new tests for `metadata_search` and `get_by_document_id`

libs/astradb/tests/integration_tests/test_vectorstore.py
```diff
@@ -818,6 +856,16 @@ async def test_astradb_vectorstore_massive_insert_replace_async(
...
+        expected_embedding_by_id = dict(zip(all_ids, all_embeddings))
+        full_results_with_embeddings = (
+            await vector_store_d2.asimilarity_search_with_embedding_id_by_vector(
+                embedding=[1.0, 1.0],
+                k=full_size,
+            )
+        )
+        for doc, embedding, doc_id in full_results_with_embeddings:
+            assert doc.page_content == expected_text_by_id[doc_id]
+            assert embedding == expected_embedding_by_id[doc_id]
```
epinzur, line 868: test of new `asimilarity_search_with_embedding_id_by_vector` method used in graph mmr

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -167,79 +244,211 @@ def __init__(
...
+            # determine if error is because of a un-indexed column. Ref:
+            # https://docs.datastax.com/en/astra-db-serverless/api-reference/collections.html#considerations-for-selective-indexing
+            error_message = str(exp).lower()
+            if ("unindexed filter path" in error_message) or (
+                "incompatible with the requested indexing policy" in error_message
+            ):
+                msg = (
+                    "The collection configuration is incompatible with vector graph "
+                    "store. Please create a new collection and make sure the path "
+                    f"`{self.metadata_incoming_links_key}` is not excluded by indexing."
```
hemidactylus, line 355: To a user, I guess it would be better to say

```
f"`metadata.{self.metadata_incoming_links_key}` is not excluded by indexing."
```

(I mean that, as a _path_ in the Data API sense, it should be spelled in full.)
One objection however is that if this store is done with autodetect, and if a flat doc schema is detected, then this path really has no `metadata.` prefix.
So perhaps the best thing here is to just say "... make sure the _metadata_ path ... is not excluded ...", just to put the reader on the right track [1]. Wdyt?

[1] The rationale here is the assumption that a reader knowledgeable enough to be dealing with a custom indexing is able to figure the details out.

libs/astradb/tests/unit_tests/test_callers.py
hemidactylus: This response is the one for the test search, no? You should be able to remove the entire `"status": ...` portion then.

libs/astradb/tests/unit_tests/test_callers.py
hemidactylus: The `"data": ...` part is what the search needs to function, so this must be kept in this mocked response.

hemidactylus (approved):
LGTM - I left a couple of minor nonblocking comments, that's all.

PR #98, opened 2024-10-22
Add Vectorize support to AstraDBGraphVectorStore

epinzur:


libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -39,13 +39,14 @@
 logger = logging.getLogger(__name__)
 
 
-class AdjacentNode:
+class EmbeddedNode:
```
epinzur, line 42: renamed so I can use it in two placees

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -262,7 +263,6 @@ def __init__(
             :meth:`~add_texts` and :meth:`~add_documents` as well.
         """
         self.metadata_incoming_links_key = metadata_incoming_links_key
-        self.embedding = embedding
```
epinzur, line 265: all embedding is now done by the underlying vector store

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -454,13 +454,20 @@ async def aadd_nodes(
...
         **kwargs: Any,
     ) -> AstraDBGraphVectorStore:
         """Return AstraDBGraphVectorStore initialized from texts and embeddings."""
-        store = cls(embedding=embedding, **kwargs)
+        store = cls(
+            embedding=embedding,
+            collection_vector_service_options=collection_vector_service_options,
+            collection_embedding_api_key=collection_embedding_api_key,
+            **kwargs,
+        )
```
epinzur, line 470: support from_text and from_documents with vectorize

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -717,21 +731,43 @@ async def ammr_traversal_search(  # noqa: C901
...
             **kwargs: Additional keyword arguments.
         """
-        query_embedding = self.embedding.embed_query(query)
-        helper = MmrHelper(
-            k=k,
-            query_embedding=query_embedding,
-            lambda_mult=lambda_mult,
-            score_threshold=score_threshold,
-        )
-
```
epinzur, line 727: move the instantiation of the helper lower down, so that we can be sure we already have the query_embedding when using vectorize

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -717,21 +731,43 @@ async def ammr_traversal_search(  # noqa: C901
...
-        async def fetch_neighborhood(neighborhood: Sequence[str]) -> None:
+        def get_candidates(nodes: Iterable[EmbeddedNode]) -> dict[str, list[float]]:
+            nonlocal outgoing_links_map
+
+            candidates: dict[str, list[float]] = {}
+            for node in nodes:
+                if node.id not in outgoing_links_map:
+                    outgoing_links_map[node.id] = _outgoing_links(node=node)
+                    candidates[node.id] = node.embedding
+            return candidates
```
epinzur, line 748: commonize some code

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -717,21 +731,43 @@ async def ammr_traversal_search(  # noqa: C901
...
+            nonlocal retrieved_docs
+
+            query_embedding, initial_nodes = await self._get_initial(
+                query=query,
+                retrieved_docs=retrieved_docs,
+                fetch_k=fetch_k,
+                filter=filter,
+            )
+
+            return query_embedding, get_candidates(nodes=initial_nodes)
```
epinzur, line 766: `fetch_initial_candidates` is moved up from just below, uses the new `self._get_initial()` (defined way below), and the new `get_candidates()` from above.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -717,21 +731,43 @@ async def ammr_traversal_search(  # noqa: C901
...
+
+        async def fetch_neighborhood_candidates(
+            neighborhood: Sequence[str],
+        ) -> dict[str, list[float]]:
```
epinzur, line 770: renamed from `fetch_neighborhood()`.  Now returns candidates using the shared `get_candidates()` method.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -753,41 +789,20 @@ async def fetch_neighborhood(neighborhood: Sequence[str]) -> None:
...
-                    outgoing_links_map[doc_id] = _outgoing_links(node=node)
-                    candidates[doc_id] = embedding
-            helper.add_candidates(candidates)
+        query_embedding, initial_candidates = await fetch_initial_candidates()
+        helper = MmrHelper(
+            k=k,
+            query_embedding=query_embedding,
+            lambda_mult=lambda_mult,
+            score_threshold=score_threshold,
+        )
```
epinzur, line 800: instantiate the helper now that we have the query_embedding.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -753,41 +789,20 @@ async def fetch_neighborhood(neighborhood: Sequence[str]) -> None:
...
+        helper.add_candidates(candidates=initial_candidates)
 
         if initial_roots:
-            await fetch_neighborhood(initial_roots)
-        if fetch_k > 0:
-            await fetch_initial_candidates()
+            neighborhood_candidates = await fetch_neighborhood_candidates(initial_roots)
+            helper.add_candidates(candidates=neighborhood_candidates)
```
epinzur, line 805: add initial_candidates and neighborhood_candidates if they exist.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -1142,14 +1157,38 @@ async def _get_outgoing_links(self, source_ids: Iterable[str]) -> set[Link]:
...
+        )
+
+        initial_nodes: list[EmbeddedNode] = []
+        for doc, embedding in result:
+            if doc.id is not None:
+                retrieved_docs[doc.id] = doc
+            initial_nodes.append(EmbeddedNode(doc=doc, embedding=embedding))
+
+        return query_embedding, initial_nodes
+
```
epinzur, line 1183: new method, using code pulled from above.  

This and `_get_adjacent()` below are the only two methods that call the `asimilarity_search_with_embedding()` method on the underlying vector store which provides the functionality for Graph MMR.

libs/astradb/langchain_astradb/graph_vectorstores.py
```diff
@@ -1172,22 +1211,23 @@ async def _get_adjacent(
...
-                    retrieved_docs[doc_id] = doc
-                if doc_id not in targets:
-                    node = _doc_to_node(doc=doc)
-                    targets[doc_id] = AdjacentNode(node=node, embedding=embedding)
+        for _, result in results:
+            for doc, embedding in result:
+                if doc.id is not None:
+                    retrieved_docs[doc.id] = doc
+                    if doc.id not in targets:
+                        targets[doc.id] = EmbeddedNode(doc=doc, embedding=embedding)
```
epinzur, line 1230: revised to handle the new response type (but ignore the query_embedding return)

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1787,43 +1787,78 @@ async def asimilarity_search_with_score_id_by_vector(
             filter=filter,
         )
 
-    async def asimilarity_search_with_embedding_id_by_vector(
+    async def asimilarity_search_with_embedding(
```
epinzur, line 1790: updated method to work with vectorize or standard vector store

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1787,43 +1787,78 @@ async def asimilarity_search_with_score_id_by_vector(
...
         self,
-        embedding: list[float],
+        query_or_embedding: str | list[float],
```
epinzur, line 1792: I first started with this as 2 different Optional params:
```
query: str | None
query_embedding: list[float] | None
```
But I was getting MANY mypy issues around missing None checks. The code got hard to follow after adding all the required checks.

So I switched to this version, and the code is easier to parse.

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1787,43 +1787,78 @@ async def asimilarity_search_with_score_id_by_vector(
...
+        """Returns the embedded query and docs most similar to query.
 
         Args:
-            embedding: Embedding to look up documents similar to.
+            query_or_embedding: Query or Embedding to look up
+                documents similar to.
             k: Number of Documents to return. Defaults to 4.
             filter: Filter on the metadata to apply.
 
+        Notes:
```
epinzur, line 1804: Should this be `Notes:` or some other heading?

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1787,43 +1787,78 @@ async def asimilarity_search_with_score_id_by_vector(
...
+
+        sort: dict[str, Any] = {}
+        include_sort_vector: bool = False
+        query_embedding: list[float] = []
+
+        if isinstance(query_or_embedding, str):
+            query: str = query_or_embedding
+            if self.document_codec.server_side_embeddings:
+                include_sort_vector = True
+                sort = {"$vectorize": query}
```
epinzur, line 1820: if query passed, and server-side embeddings, then:

* use vectorize
* return embedding on query

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1787,43 +1787,78 @@ async def asimilarity_search_with_score_id_by_vector(
...
+            else:
+                query_embedding = self._get_safe_embedding().embed_query(text=query)
+                sort = self.document_codec.encode_vector_sort(vector=query_embedding)
```
epinzur, line 1823: if query passed, and client-side embeddings, then:
* embed the query
* query by vector

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1787,43 +1787,78 @@ async def asimilarity_search_with_score_id_by_vector(
...
-                continue
+            include_sort_vector=include_sort_vector,
+            sort=sort,
+        )
+        if include_sort_vector:
+            sort_vector = await async_cursor.get_sort_vector()
+            if sort_vector is None:
+                msg = "Unable to retrieve the server-side embedding of the query."
+                raise ValueError(msg)
+            query_embedding = sort_vector
```
epinzur, line 1846: if using vectorize to embed the query, pull the embedding

libs/astradb/tests/integration_tests/conftest.py
```diff
@@ -464,7 +464,7 @@ def collection_idxall_vz(
     """
     collection = database.create_collection(
         COLLECTION_NAME_IDXALL_VZ,
-        dimension=16,
+        dimension=1536,
```
epinzur, line 467: I needed to boost these to the full size to get some of my tests to pass. Is that okay?

Note: I didn't try any sizes in-between.  256 or something may work.  I need enough context to tell the difference between the text groupings in these documents:

https://github.com/langchain-ai/langchain-datastax/pull/98/files#diff-e8f42cd690704e0cec0df50a7389d5a2f0fc09ba0f8b1b91839601dc94509b53R109-R142

libs/astradb/tests/integration_tests/test_graphvectorstore.py
```diff
@@ -5,7 +5,7 @@
```
epinzur, line 5: parametrized existing tests with vectorize setups.

libs/astradb/tests/integration_tests/test_vectorstore.py
```diff
@@ -856,16 +856,6 @@ async def test_astradb_vectorstore_massive_insert_replace_async(
...
-        expected_embedding_by_id = dict(zip(all_ids, all_embeddings))
-        full_results_with_embeddings = (
-            await vector_store_d2.asimilarity_search_with_embedding_id_by_vector(
-                embedding=[1.0, 1.0],
-                k=full_size,
-            )
-        )
-        for doc, embedding, doc_id in full_results_with_embeddings:
-            assert doc.page_content == expected_text_by_id[doc_id]
-            assert embedding == expected_embedding_by_id[doc_id]
```
epinzur, line 868: moved the test of this to its own block below, where I can test more combinations of setups

libs/astradb/tests/integration_tests/test_vectorstore.py
```diff
@@ -1419,6 +1409,73 @@ async def test_astradb_vectorstore_similarity_scale_async(
...
+                    query_or_embedding={"test": "error"}  # type: ignore # noqa: PGH003
+                )
+            return
+
+        assert_list_of_numeric(query_embedding)
+        assert isinstance(results, list)
+        assert len(results) > 0
+        (doc, embedding) = results[0]
+        assert isinstance(doc, Document)
+        assert_list_of_numeric(embedding)
```
epinzur, line 1477: assert shape, not content, so works across different vector stores and embeddings.  GVS MMR tests cover actual performance (similarity results) of this code in more detail.

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: I have two questions (not blockers but just to clarify the choices)

1. This flow is async-only (as opposed to most of the rest of the vector store). I see the sync counterpart (in the GVS) wraps an asyncio.run call to this one. Could there be some use where no event loop is available or something? (I am genuinely asking here, not sure if this is a real possibility)
2. The double-argument implementation surely led to a forest of None-checks (bad). This one, however, relies on `isinstance` checks, which are to be avoided when possible. Have you considered splitting like in other places of this class, `asimilarity_search_by_sort` and two thin entry-functions, one specific for query:str and the other for embedding:list[float] ? The former would contain the "if vectorize" and they would converge on passing their proper `sort={...}` to the `asimilarity_search_by_sort`. This would probably be cleaner (it is the philosophy outlined in the chart in [this PR message](https://github.com/langchain-ai/langchain-datastax/pull/57#issue-2484873779)).

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: Suggestion: always include_sort_vector, always use the one read from the response for the subsequent MMR. Small waste of bandwidth, nice simplification of code.

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: see comment above on always using the include_s_v and reading it afterwards (it works with `{sort: {$vector: ...}}` all right.

libs/astradb/tests/integration_tests/conftest.py
hemidactylus: No objections from me (this should not affect the test robustness AFAIK)

libs/astradb/tests/integration_tests/test_graphvectorstore.py
```diff
@@ -579,21 +909,20 @@ async def test_gvs_add_nodes_async(
...
-            Node(text="[0, 1]", metadata={"m": 1}, links=links1),
+            Node(id="id0", text=page_contents[0], metadata={"m": 0}, links=links0),
+            Node(text=page_contents[1], metadata={"m": 1}, links=links1),
         ]
-        async for _ in graph_vector_store_d2.aadd_nodes(nodes):
+        async for _ in g_store.aadd_nodes(nodes):
             pass
 
-        hits = await graph_vector_store_d2.asimilarity_search_by_vector([0, 3])
+        query = "italian food" if is_vectorize else "[0, 3]"
```
hemidactylus, line 918: "italian food" :heart:

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: (I see this may be outdated by now). I _think_ Google docstring conventions mandate a `Note:` (singular) heading, possibly repeated over one paragraph for each note. At least this is what I did [here](https://github.com/datastax/astrapy/blob/ab67b6167bf71d9703b8afd0fe96da24433d8d20/astrapy/data/collection.py#L714-L719) for example - rendered as you see [here](https://docs.datastax.com/en/astra-api-docs/_attachments/python-client/astrapy/collection.html#astrapy.collection.Collection.insert_many).

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: I checked with @cbornet on point 1.
This pattern (wrapping the async in the sync version) is indeed discouraged -- this code may not have control over the event loop, so this may have unpredictable failure modes (e.g. within a notebook it could well fail).
The suggestion is to implement the sync method (the GVS' `mmr_traversal_search`, I mean) independently of the async one - possibly refactoring the parts that do not make requests into common code.

hemidactylus: Also, if not going with the clean split outlined above, this change to `mmr_traversal_search` would make it a bit more robust:

```
    @override
    def mmr_traversal_search(
        self,
        ...
    ) -> Iterable[Document]:

        ...

        try:
            loop = asyncio.get_running_loop()
        except RuntimeError:
            return asyncio.run(collect_docs())
        return loop.run_until_complete(collect_docs())
```

epinzur: Going to do item 1 in a different PR. Item 2 is implemented in the latest commit.

hemidactylus (approved):
LGTM, thank you for your attention to everything!

PR #100, opened 2024-10-24
Create 'real' synchronous methods for GVS

epinzur:
* Removed wrapping async methods with asyncio.run() for their synchronous counterparts.  
* Created actual synchronous methods instead.  Existing synchronous tests pass without changes.
* Also created asynchronous versions of 2 methods that were previously missing.


hemidactylus (approved):
LGTM

PR #105, opened 2024-11-26
Use blockbuster to detect blocking calls in the event loop

cbornet:


hemidactylus (approved):
LGTM (tested locally to overcome the puzzling env var thing and all passed. Running on commit `6452940fb49d2074646ca68b741869e24afd5624`).

PR #111, opened 2025-02-03
Add "with_component_name" method to AstraDBVectorStore

hemidactylus:
This method returns a copy of the vector store with all properties unchanged except the caller information (used for usage stats by Astra DB's Data API), where the "component name" is replaced by a new supplied value.

This is needed by the implementation of Graph Vector Store as a Retriever delegating to the vector store class.

The method does instantiate a new AstraDBVectorStore with dummy values and SetupMode=OFF to avoid triggering any actual work, then proceeds to replace all "private" members of the new instance to make it an exact copy, save for the "component name". This has required a similar logic to be coded into the AstraDBEnvironment/CollectionEnvironment as well.
(The alternative of making room in the constructor for such a path was deemed too consequence-bearing to be pursued.)

### DIFF
--- a/libs/astradb/langchain_astradb/utils/astradb.py
+++ b/libs/astradb/langchain_astradb/utils/astradb.py
@@ -283,18 +283,20 @@ class _AstraDBEnvironment:
         # - a single ("langchain", <version of langchain_core>)
         # - if such is provided, a (component_name, <version of langchain_astradb>)
         #   (note: if component_name is None, astrapy strips it out automatically)
+        self.ext_callers = ext_callers
+        self.component_name = component_name
         norm_ext_callers = [
             cpair
             for cpair in (
                 _raw_caller if isinstance(_raw_caller, tuple) else (_raw_caller, None)
-                for _raw_caller in (ext_callers or [])
+                for _raw_caller in (self.ext_callers or [])
             )
             if cpair[0] is not None or cpair[1] is not None
         ]
         full_callers = [
             *norm_ext_callers,
             LC_CORE_CALLER,
-            (component_name, LC_ASTRADB_VERSION),
+            (self.component_name, LC_ASTRADB_VERSION),
         ]
 
         # create the callers
@@ -343,9 +345,10 @@ class _AstraDBCollectionEnvironment(_AstraDBEnvironment):
             async_astra_db_client=async_astra_db_client,
         )
         self.collection_name = collection_name
+        self.collection_embedding_api_key = collection_embedding_api_key
         self.collection = self.database.get_collection(
             name=self.collection_name,
-            embedding_api_key=collection_embedding_api_key,
+            embedding_api_key=self.collection_embedding_api_key,
         )
         self.async_collection = self.collection.to_async()
 
@@ -394,6 +397,24 @@ class _AstraDBCollectionEnvironment(_AstraDBEnvironment):
                         raise data_api_exception  # noqa: TRY201
                 except ValueError as validation_error:
                     raise validation_error from data_api_exception
+
+    def with_component_name(self, component_name: str) -> _AstraDBCollectionEnvironment:
+        """Create a copy of this environment with just the 'component name' changed.
+
+        Attributes:
+            component_name: the new value, which replaces the current one in the copy.
+        """
+        return _AstraDBCollectionEnvironment(
+            collection_name=self.collection_name,
+            token=self.token,
+            api_endpoint=self.api_endpoint,
+            keyspace=self.keyspace,
+            environment=self.environment,
+            ext_callers=self.ext_callers,
+            component_name=component_name,
+            setup_mode=SetupMode.OFF,
+            collection_embedding_api_key=self.collection_embedding_api_key,
+        )
 
     async def _asetup_db(
         self,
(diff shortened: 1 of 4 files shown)


libs/astradb/langchain_astradb/utils/astradb.py
```diff
@@ -395,6 +398,24 @@ def __init__(
                 except ValueError as validation_error:
                     raise validation_error from data_api_exception
 
+    def with_component_name(self, component_name: str) -> _AstraDBCollectionEnvironment:
```
cbornet, line 401: How about having a `copy(**kwargs)` method ?

```python
def copy(**kwargs):
    return _AstraDBCollectionEnvironment(
            collection_name=kwargs.get("collection_name", self.collection_name),
            token=kwargs.get("token", self.token)
            api_endpoint=kwargs.get("api_endpoint", self.api_endpoint),
            keyspace=kwargs.get("keyspace", self.keyspace),
            environment=kwargs.get("environment", self.environment),
            ext_callers=kwargs.get("ext_callers", self.ext_callers),
            component_name=kwargs.get("component_name", self.component_name),
            setup_mode=kwargs.get("setup_mode", SetupMode.OFF),
            collection_embedding_api_key=kwargs.get("collection_embedding_api_key", self.collection_embedding_api_key),
        )
```

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -723,6 +724,45 @@ def _select_relevance_score_fn(self) -> Callable[[float], float]:
         # so here the final score transformation is not reversing the interval.
         return lambda score: score
 
+    def with_component_name(self, component_name: str) -> AstraDBVectorStore:
```
cbornet, line 727: Could also be `copy(**kwargs)`

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -723,6 +724,45 @@ def _select_relevance_score_fn(self) -> Callable[[float], float]:
...
+        """
+        copy = AstraDBVectorStore(
+            collection_name="moot",
+            api_endpoint="http://moot",
+            environment=Environment.OTHER,
+            namespace="moot",
+            setup_mode=SetupMode.OFF,
+            collection_vector_service_options=CollectionVectorServiceOptions(
+                provider="moot",
+                model_name="moot",
```
cbornet, line 741: Why not passing the final values instead of `moot` directly ?

hemidactylus: My intent here was to ensure future readers understand the nature of this "trick", that the instantiation is misused so to speak - exploiting a path in the constructor where nothing happens.

The concern was that passing the real values (which however need to be supplied with more member-setting right thereafter) might mislead readers.

libs/astradb/langchain_astradb/utils/astradb.py
hemidactylus: The reason I did not generalize so much is that not many arguments (to the constructor) would lean correctly to this kind of "duplication with just some tweaks". For instance:
- we hardcode setupMode to be OFF since we want this "copy" to not do any side-effect and be instantaneous (this is the main obstacle)
- it would be strange, then, to expose the possibility of changing collection name since we don't offer a way to go through the usual create-collection flow. (For such operations, I suppose, we should only leave the "regular legitimate constructor" route)
- ditto for api endpoint, environment

In my opinion, perhaps the case could be made for this "copy" to support on-the-fly change of just `token`, `ext_callers`, `collection_embedding_api_key` and `component_name`, probably nothing else would make it "safe" for usage by the coder from the outside.

But then, calling it "copy" seems a stretch and it could remain as is (possibly generalized to these 4 params and aliased if ever the need arose).

At least that was my reasoning, I'm happy to discuss it further!

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: (same reasoning as above)

libs/astradb/langchain_astradb/utils/astradb.py
cbornet: I see. I would still lean on calling it copy() and limiting the args that can be changed.
If you call `copy()` without args, that's effectively a shallow copy.
The question is then how to signal to the user which args are allowed.
Maybe it can be in the signature with meaning `None` is take value from original and not set it to None ?
I originally proposed kwargs because it could distinguish between "not set" and "set to None" but maybe "set to None" is not needed/allowed.

hemidactylus: This is a valid point. I am renaming to `copy` right now, with a modicum of docstring to highlight the details of behaviour.
Also the eternal "unset VS None" issue in this case does not bite since I agree that passing explicit None has no purpose whatsoever (esp. considering there are idiomatic ways to suppress token and/or API Key that do not involve passing None. I mention them in the new docstring).

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: I'm adapting this in line with above. ==> `copy`

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: > exploiting a path in the constructor where nothing happens.
I am not very proud of this solution, to be honest. But it probably seems overkill to refactor everything else to make this little flow more straightforward.

cbornet (approved):
LGTM!

PR #113, opened 2025-02-13
Vector Store, full separation codec / vectorstore

### LINKED ISSUE #106: Inconsistency between codec and metadata_indexing
It seems like `metadata_indexing_{include,exclude}` may not work depending on the codec.

Specifically, `metadata_indexing_include = ["foo"]` is always re-written to exclude `metadata.foo`, [1] but the Flat codec will write `metadata["foo"]` to foo  [2].

[1]: https://github.com/langchain-ai/langchain-datastax/blob/main/libs/astradb/langchain_astradb/vectorstores.py#L345
[2]: https://github.com/langchain-ai/langchain-datastax/blob/main/libs/astradb/langchain_astradb/utils/vector_store_codecs.py#L366

hemidactylus:
Clean separation of codecs/vectorstore layers

This PR introduces a better separation of knowledge on what pertains to codecs and what constitutes the underlying logic of the Vector Store. As such,

fixes #106 

Meanwhile, it provides a couple of useful querying tools which feel like they belong to the vstore+codec layer. (though another such tool, the "run_query" method, is postponed to a follow-up PR. That one is possibly the most important of these.)

Additionally, this restructuring of the code also makes a step toward a possible extension to cover API Tables without duplicating logic.

More in detail:

- creation of "id queries" in the codec consistently (away from vectorstore)
- moved id- and $vector-related parts of codec into the base class (not expected to vary under the data api)
- codecs expose their "default indexing policy", vectorstore uses that knowledge in and around its constructor ( --> for coll. creation in particular)
- codecs expose their "abstract metadata key to actual dot-notation field identifier" for building include/exclude policies specified by _metadata_ fields
- vectorstore does not say "_id" literally anymore (all is in codec; uses get_id)

A **note** on the name chosen for the `get_id` codec method. There is the unfortunate fact that LangChain calls "documents" its internal format, and Astra DB calls "document" what it stores. So ... codecs that lie between these two have sometimes a hard time with naming. Keeping it simple (`get_id`) should work in this case because only one of the two ends of the codec (the Astra side) can have a variable schema: then, only on that side should the need arise for a function abstracting the reading of an ID. (Which btw is more a formality as the `"_id"` field is one of those that will hardly ever change in Astra DB!)

### DIFF
--- a/libs/astradb/langchain_astradb/utils/vector_store_codecs.py
+++ b/libs/astradb/langchain_astradb/utils/vector_store_codecs.py
@@ -16,11 +16,21 @@ VECTOR_REQUIRED_PREAMBLE_MSG = (
 )
 FLATTEN_CONFLICT_MSG = "Cannot flatten metadata: field name overlap for '{field}'."
 
+STANDARD_INDEXING_OPTIONS_DEFAULT = {"allow": ["metadata"]}
+
 logger = logging.getLogger(__name__)
 
 
 def _default_decode_vector(astra_doc: dict[str, Any]) -> list[float] | None:
     return astra_doc.get("$vector")
+
+
+def _default_metadata_key_to_field_identifier(md_key: str) -> str:
+    return f"metadata.{md_key}"
+
+
+def _flat_metadata_key_to_field_identifier(md_key: str) -> str:
+    return md_key
 
 
 def _default_encode_filter(filter_dict: dict[str, Any]) -> dict[str, Any]:
@@ -37,7 +47,7 @@ def _default_encode_filter(filter_dict: dict[str, Any]) -> dict[str, Any]:
                 # assume each list item can be fed back to this function
                 metadata_filter[k] = _default_encode_filter(v)  # type: ignore[assignment]
         else:
-            metadata_filter[f"metadata.{k}"] = v
+            metadata_filter[_default_metadata_key_to_field_identifier(k)] = v
 
     return metadata_filter
 
@@ -136,6 +146,26 @@ class _AstraDBVectorStoreDocumentCodec(ABC):
         Returns:
             an equivalent filter clause for use in Astra DB's find queries.
         """
+
+    @abstractmethod
+    def metadata_key_to_field_identifier(self, md_key: str) -> str:
+        """Express an 'abstract' metadata key as a full Data API field identifier."""
+
+    @property
+    @abstractmethod
+    def default_collection_indexing_policy(self) -> dict[str, list[str]]:
+        """Provide the default indexing policy if the collection must be created."""
+
+    def get_id(self, astra_document: dict[str, Any]) -> str:
+        """Return the ID of an encoded document (= a raw JSON read from DB)."""
+        return astra_document["_id"]
+
+    def get_similarity(self, astra_document: dict[str, Any]) -> float:
+        """Return the similarity of an encoded document (= a raw JSON read from DB).
+
+        This method assumes its argument comes from a suitable vector search.
+        """
+        return astra_document["$similarity"]
 
     def encode_id(self, filter_id: str) -> dict[str, Any]:
         """Encode an ID as a filter for use in Astra DB queries.
@@ -244,6 +274,14 @@ class _DefaultVSDocumentCodec(_AstraDBVectorStoreDocumentCodec):
     def encode_filter(self, filter_dict: dict[str, Any]) -> dict[str, Any]:
         return _default_encode_filter(filter_dict)
 
+    @override
+    def metadata_key_to_field_identifier(self, md_key: str) -> str:
+        return _default_metadata_key_to_field_identifier(md_key)
+
+    @property
+    def default_collection_indexing_policy(self) -> dict[str, list[str]]:
+        return STANDARD_INDEXING_OPTIONS_DEFAULT
+
 
 class _DefaultVectorizeVSDocumentCodec(_AstraDBVectorStoreDocumentCodec):
     """Codec for the default vector store usage with server-side embeddings.
@@ -318,6 +356,14 @@ class _DefaultVectorizeVSDocumentCodec(_AstraDBVectorStoreDocumentCodec):
     @override
     def encode_filter(self, filter_dict: dict[str, Any]) -> dict[str, Any]:
         return _default_encode_filter(filter_dict)
+
+    @property
+    def default_collection_indexing_policy(self) -> dict[str, list[str]]:
+        return STANDARD_INDEXING_OPTIONS_DEFAULT
+
+    @override
+    def metadata_key_to_field_identifier(self, md_key: str) -> str:
+        return _default_metadata_key_to_field_identifier(md_key)
 
 
 class _FlatVSDocumentCodec(_AstraDBVectorStoreDocumentCodec):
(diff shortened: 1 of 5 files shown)


libs/astradb/langchain_astradb/utils/vector_store_codecs.py
```diff
@@ -132,17 +148,49 @@ def encode_filter(self, filter_dict: dict[str, Any]) -> dict[str, Any]:
...
         Args:
             filter_id: the ID value to filter on.
 
         Returns:
-            an filter clause for use in Astra DB's find queries.
+            a filter clause for use in Astra DB's find queries.
         """
+        return _default_encode_id(filter_id)
+
+    def encode_ids(self, filter_ids: list[str]) -> dict[str, Any]:
```
bjchambers, line 181: Two potential issues with this being a separate method:

1. For a user that has a query for "these IDs AND this predicate" they're going to need to write the `_id` filter themselves unless they have access to the codec.
2. We'd also need the rewriting that happens to special case the `_id` filter and not rewrite it to `metadata._id`.

I wonder if we should just adopt a `$id` or `_id` as the standar field for the `id`. Then, I don't know if we'd need the `encode_ids` -- we could just make the rewriter do the right thing in that case, and the user can provide `{ "$id": <id> }` or `{ "$id": { "$in": [<ids>] }}` depending on their needs in arbitrary places within the query.

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1333,10 +1313,11 @@ async def _replace_document(
                 document: dict[str, Any],
             ) -> tuple[UpdateResult, str]:
                 async with sem:
+                    doc_id = self.document_codec.get_id(document)
                     return await _async_collection.replace_one(
-                        {"_id": document["_id"]},
+                        self.document_codec.encode_id(doc_id),
```
bjchambers, line 1318: The idea of allowing the codec to just be a filter rewriter actually would be pretty useful here as well. It would allow these to just be "do a query with `codec.rewrite_filters({"$id": doc_id})` (don't remember the name it currently has) which feels like it's clearer what is being done. Then instead of having a bunch of methods on the codec that are used for creating queries, there is just the one that converts a "standard" query (on the `Document` basically) into a appropriately encoded query.

hemidactylus:
@bjchambers Thank you for taking the time to review this. You make good points.
And you are right, a refinement of this separation is in order.

I have added the "do not merge" label while I go over it again. Here are the two points I take from your remarks.

- I would still keep the codec unaware of logic such as "these filters are too many, let's split" (this does not belog there)
- but I agree the codec should expose a more general method, which just translates "any query"

More on the second point: "any query", in LC world -- where `Document` objects have just: (1) `id`, (2) a vector, (3) metadata kv pairs and (4) an unindexed text -- means that the most general thing to query is:
- zero, one or more IDs (if more: implied OR)
- a search vector with its `k` for ANN
- metadata conditions. These are: `k: [v1, v2...]` ==> implied OR, but also `{k1:v1, k2:v2...}` where it can be AND/OR equally. We could assume AND between different keys (do we worry about loss of generality on this?) or we can keep a Data-API-like syntax here.

What I'm trying to get to is, since the codec's "query translator" makes an abstract query into a payload good to go to Data API, its input could be required to be _not a dictionary_ (which arguably makes usage more error-prone), rather a specific structure: instances of some `AstraDBDocumentQuery` class, to be then made into dictionaries only with the knowledge of the encoding scheme.

I can probably find some time to rework this PR in this sense tomorrow - would that capture the essence of your remarks? (plus helping with clarity since it requires data classes to express the abstractness of queries? Or perhaps too unwieldy?)

bjchambers:
I like that direction and the observation it doesn't all need to be in a dict. That will keep the difference between `id` and `metadata["id"]` clear too.

I wonder if a data class is necessary. Could it just be:

```
def encode_query(self, ids: Iterable[str | int] = (), metadata: dict[str, Any] = {}):
```

I think keeping vector separate (it goes to sort, not the filter) could be reasonable.

hemidactylus:
I have replaced the "encode_id[s]" for a `encode_query` along the lines we discussed.

Some notes:

- `encode_query` assumes the IDs are in `AND` with the metadata conditions (if both passed). (I think unlikely one wants and OR between those - in which case, running multiple queries and merging the results would be the way to go I believe)
- the second parameter is named `filter_dict` for compatibility with the name throughout the `VectorStore` class, where it always means "metadata filters"
- `ids` needs not be typed as `str | int`. In Langchain IDs are always strings I believe

I have adapted the unit tests (`test_vs_doc_codecs.py`, tests `test_flat/default_query_encoding`). Below a summary of the wonders of `encode_query` for your convenience:

```python
from langchain_astradb.utils.vector_store_codecs import _DefaultVSDocumentCodec, _FlatVSDocumentCodec
d = _DefaultVSDocumentCodec(content_field='c', ignore_invalid_documents=False)
f = _FlatVSDocumentCodec(content_field='c', ignore_invalid_documents=False)


d.encode_query()
f.encode_query()
# both: {}


d.encode_query(ids=['id1'])
f.encode_query(ids=['id1'])
# both: {'_id': 'id1'}


d.encode_query(ids=['id1', 'id2'])
f.encode_query(ids=['id1', 'id2'])
# both: {'_id': {'$in': ['id1', 'id2']}}


d.encode_query(ids=['d'],filter_dict={'x':'y'})
f.encode_query(ids=['d'],filter_dict={'x':'y'})
# resp.:
#   {'$and': [{'_id': 'd'}, {'metadata.x': 'y'}]}
#   {'$and': [{'_id': 'd'}, {'x': 'y'}]}


d.encode_query(ids=['d'],filter_dict={'x':'y','z':'w'})
f.encode_query(ids=['d'],filter_dict={'x':'y','z':'w'})
# resp.:
#   {'$and': [{'_id': 'd'}, {'metadata.x': 'y', 'metadata.z': 'w'}]}
#   {'$and': [{'_id': 'd'}, {'x': 'y', 'z': 'w'}]}


d.encode_query(ids=['d'],filter_dict={'$or':[{'x':'y'},{'z':'w'}]})
f.encode_query(ids=['d'],filter_dict={'$or':[{'x':'y'},{'z':'w'}]})
# resp.:
#   {'$and': [{'_id': 'd'}, {'$or': [{'metadata.x': 'y'}, {'metadata.z': 'w'}]}]}
#   {'$and': [{'_id': 'd'}, {'$or': [{'x': 'y'}, {'z': 'w'}]}]}
```

hemidactylus:
> the difference between `id` and `metadata["id"]`

(an incidental note is that trying to use `metadata._id` as a literal metadata field may stop working as soon as one has a "flat" vector store - for reasons unrelated to this PR, more fundamental. I believe however that the code should not throw an error - suppose there is a legacy vectorstore out there that uses such a metadata field, created prior to the introduction of flat codecs and autodetect. Certainly not something to encourage, ...)

libs/astradb/langchain_astradb/utils/vector_store_codecs.py
```diff
@@ -132,17 +148,44 @@ def encode_filter(self, filter_dict: dict[str, Any]) -> dict[str, Any]:
...
+        This method assumes its argument comes from a suitable vector search.
         """
+        return astra_document["$similarity"]
+
+    def encode_query(
+        self,
+        *,
+        ids: Iterable[str] | None = None,
+        filter_dict: dict[str, Any] | None = None,
+    ) -> dict[str, Any]:
```
bjchambers, line 175: May merit pydoc indicating the implicit `$and`.

libs/astradb/langchain_astradb/utils/vector_store_codecs.py
```diff
@@ -132,17 +148,44 @@ def encode_filter(self, filter_dict: dict[str, Any]) -> dict[str, Any]:
...
+        clauses: list[dict[str, Any]] = []
+        _ids_list = list(ids or [])
+        if _ids_list:
+            clauses.append(_default_encode_ids(_ids_list))
+        if filter_dict:
+            clauses.append(self.encode_filter(filter_dict))
+
+        if clauses:
+            if len(clauses) > 1:
+                return {"$and": clauses}
```
bjchambers, line 185: Agreed this makes sense. In general, my rationale is:

1. If you want `ids OR filter`, then run separate queries -- the IDs only query is likely to be fast and the filter-only query is likely to do a scan.
2. If you want `ids AND filter` then there is no option beyond running them together (unless you somehow emulate the full filtering semantics on the client side).

So, it seems like the `AND` is the only reasonable choice.

bjchambers (approved):

PR #114, opened 2025-02-21
Vector store's run_query method

hemidactylus:
This PR introduces a method `run_query` (+async equivalent) for the AstraDBVectorStore.

The method supports a variety of call patterns (ANN/non-ANN sorting, filters on ID and metadata; boolean flags to return embeddings, query vector, similarities; including a mapper function to transform the returned items).

The return value of `run_query` follows a fixed tuple-based schema, meant to be unpacked by the caller to pick what is needed: when some of the results are not requested (e.g. no embeddings), None values are found in the result (this saves some latency as it actually avoids their server-side computation and on-wire transfer).

This method aims at exposing an all-purpose method for running queries on a vector store that has three important properties:
1. clean separation of codec concerns and query logic. Callers of `run_query` are not supposed to know anythign about the actual structure of the documents on DB (unless they want to pass a custom mapping from raw returned documents, that is). The translation passes through the codec consistently in a manner managed by `run_query`
2. allows for enough flexibility in the query pattern to be reasonably seen as the one centralized piece of logic driving all interrogations of the collection;
3. ... and as such, in this PR, it is indeed used by all other read-path methods of the vector store, which defer to it (and its usage of the codecs).

Furthermore, this PR completes the few remaining pieces of the work initiated in #113 by moving a few raw-document interactions to the codec layer (for example, reading the `["$vector"]` of a document). Even though those do no harm if hardcoded in the vector store, (e.g. the location of the vector will hardly change in Data API), clean code organization required this separation to be formally complete.

A remaining concern is whether the (async) `arun_query` should optionally accept an **awaitable** mapper function. This is relevant only if said function performs blocking side-effects in its body. For the moment I assumed this not to be the case.

### DIFF
--- a/libs/astradb/langchain_astradb/vectorstores.py
+++ b/libs/astradb/langchain_astradb/vectorstores.py
@@ -1442,6 +1442,69 @@ class AstraDBVectorStore(VectorStore):
 
         return sum(u_res.update_info["n"] for u_res in update_results)
 
+    def run_query(
+        self,
+        *,
+        n: int,
+        ids: list[str] | None = None,
+        filter: dict[str, Any] | None = None,  # noqa: A002
+        ann_sort: dict[str, Any] | None = None,
+        include_similarity: bool | None = None,
+        include_sort_vector: bool | None = None,
+        include_embeddings: bool | None = None,
+        raw_document_mapper: Callable[[dict[str, Any]], Any | None] | None = None,
+    ) -> tuple[
+        list[float] | None, Iterable[tuple[Any, str, list[float] | None, float | None]]
+    ]:
+        """Execute a generic query on stored documents.
+
+        The return value has the following shape:
+        (qvector?, [(mappeds, id, embedding?, similarity?)])
+
+        projection is dictated by include_embeddings essentially
+
+        Note: if codec skips invalid docs, you may get <k results even if
+        there were more.
+        """
+        find_query = self.document_codec.encode_query(
+            ids=ids,
+            filter_dict=filter,
+        )
+        find_sort = ann_sort
+        find_projection = (
+            self.document_codec.full_projection
+            if include_embeddings
+            else self.document_codec.base_projection
+        )
+        mapper: Callable[[dict[str, Any]], Any | None]
+        if raw_document_mapper:
+            mapper = raw_document_mapper
+        else:
+            mapper = self.document_codec.decode
+
+        find_raw_iterator = self.astra_env.collection.find(
+            filter=find_query,
+            projection=find_projection,
+            limit=n,
+            include_similarity=include_similarity,
+            include_sort_vector=include_sort_vector,
+            sort=find_sort,
+        )
+
+        sort_vector = find_raw_iterator.get_sort_vector()
+        final_doc_iterator = (
+            (
+                mapper(raw_doc),
+                self.document_codec.get_id(raw_doc),
+                raw_doc.get("$vector") if include_embeddings else None,
+                self.document_codec.get_similarity(raw_doc)
+                if include_similarity
+                else None,
+            )
+            for raw_doc in find_raw_iterator
+        )
+        return sort_vector, final_doc_iterator
+
     def metadata_search(
         self,
         filter: dict[str, Any] | None = None,  # noqa: A002
@@ -1687,7 +1750,7 @@ class AstraDBVectorStore(VectorStore):
                 "embeddings is not allowed."
             )
             raise ValueError(msg)
-        sort = {"$vector": embedding}
+        sort = self.document_codec.encode_vector_sort(embedding)
         return self._similarity_search_with_score_id_by_sort(
             sort=sort,
             k=k,
@@ -1883,7 +1946,7 @@ class AstraDBVectorStore(VectorStore):
                 "embeddings is not allowed."
             )
             raise ValueError(msg)
-        sort = {"$vector": embedding}
+        sort = self.document_codec.encode_vector_sort(embedding)
         return await self._asimilarity_search_with_score_id_by_sort(
             sort=sort,
             k=k,
@@ -2214,7 +2277,7 @@ class AstraDBVectorStore(VectorStore):
         metadata_parameter = self.filter_to_query(filter)
 
         return self._run_mmr_query_by_sort(
-            sort={"$vector": embedding},
+            sort=self.document_codec.encode_vector_sort(embedding),
             k=k,
             fetch_k=fetch_k,
             lambda_mult=lambda_mult,
@@ -2253,7 +2316,7 @@ class AstraDBVectorStore(VectorStore):
         metadata_parameter = self.filter_to_query(filter)
 
         return await self._arun_mmr_query_by_sort(
-            sort={"$vector": embedding},
+            sort=self.document_codec.encode_vector_sort(embedding),
             k=k,
             fetch_k=fetch_k,
             lambda_mult=lambda_mult,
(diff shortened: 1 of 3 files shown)


libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1442,6 +1447,331 @@ async def _update_document(
...
+            include_similarity: whether to return similarity scores with each match.
+                Requires vector sort.
+            include_sort_vector: whether to return the very query vector used for the
+                ANN search alongside the iterable of results. Requires vector sort.
+            include_embeddings: whether to retrieve the matches' own embedding vectors.
+            raw_document_mapper: if provided, an arbitrary function to process the raw
+                documents, as found on Astra DB, and possibly convert them into another
+                format (`V`). If passed, it must be `Callable[[dict], V | None]`.
+                If omitted, the vector store's codec is used, returning Documents.
+                When providing a custom mapper, it is the caller's responsibility to
```
bjchambers, line 1547: I wonder if it would make sense to do a layer of normalization before generating the dict, otherwise the user will need to handle the codec, right?

Eg., move everything to be something like the following fields in a named dict?

* `id`
* `content`
* `metadata`
* `embedding`
* `similarity_score`

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1442,6 +1447,331 @@ async def _update_document(
...
+        sort: dict[str, Any] | None = None,
+        include_similarity: bool | None = None,
+        include_sort_vector: bool | None = None,
+        include_embeddings: bool | None = None,
+    ) -> tuple[
+        list[float] | None,
+        Iterable[tuple[Document, str, list[float] | None, float | None]],
+    ]: ...
+
+    @overload
```
bjchambers, line 1467: It may be possible to avoid the overload if the default document mapper was `Callable[[DocDict], Document]`, right?

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1442,6 +1447,331 @@ async def _update_document(
...
+
+        Args:
+            n: amount of items to return. Fewer items than `n` may be returned in the
+                following cases: (a) the document mapper skips some raw entries from
+                the server; (b) the collection has not enough matches.
+            ids: a list of document IDs to restrict the query to. If this is supplied,
+                only document with an ID among the provided one will match. If further
+                query filters are provided (i.e. metadata), matches must satisfy both
+                requirements.
+            filter: a metadata filtering part. If provided, if must refer to
```
bjchambers, line 1528: "If provided, if" -> "If provided, it must"?

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: Probably, good idea. Let me check...

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: Yeah, I was not sure about this one but you are right.

What made me hesitate was that blocking access to the actual raw doc might block some very custom needs to access fields otherwise hidden (e.g. a non-content top-level field in a nested-codec collection), but on second thought such a case should be actively discouraged as the vector store already would not be able to see it.

So yes. Also the named dict would make it easier to extend later with a `raw_document`, should we ever need it.

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: thanks!

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: I am not sure how to avoid the overload actually. The default value (to set after the `=`, I mean) is `self.document_codec.blablabla`, so I cannot put it in the signature as a default value.

For that reason I went with a `None`, which signals "use codec conversion" to the function body. I am not sure how to handle it the way you hint at without overload.

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -56,13 +61,26 @@
...
     from astrapy.results import UpdateResult
-    from langchain_core.documents import Document
     from langchain_core.embeddings import Embeddings
 
 T = TypeVar("T")
 U = TypeVar("U")
 DocDict = Dict[str, Any]  # dicts expressing entries to insert
 
+
+class FullQueryResult(NamedTuple):
```
bjchambers, line 71: Thought on `AstraQueryResult`?

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1442,6 +1460,373 @@ async def _update_document(
...
+        return None
+
+    def run_query_raw(
+        self,
+        *,
+        n: int,
+        ids: list[str] | None = None,
+        filter: dict[str, Any] | None = None,  # noqa: A002
+        sort: dict[str, Any] | None = None,
+        include_similarity: bool | None = None,
```
bjchambers, line 1509: Why not indicate the defaults here?

`include_similarity: bool = False` rather than allowing it to be optional?

libs/astradb/langchain_astradb/vectorstores.py
```diff
@@ -1442,6 +1460,373 @@ async def _update_document(
...
+        include_sort_vector: bool | None = None,
+        include_embeddings: bool | None = None,
+    ) -> tuple[list[float] | None, Iterable[DocDict]]:
```
bjchambers, line 1512: I could see a case for an overload here for `tuple[list[float], Iterable[DocDict]] | Iterable[DocDict]` based on `include_sort_vector`. I could also see a case for "this is an advanced query method, and it's ok to be a little messy, since the advanced uses are likely to be wrapping it rather than using it directly".

bjchambers (approved):

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: The current "None" convention would ultimately defer defaults to the Data API. If the Data API ever changes its defaults (it happened once with $vector being excluded from the default projection), right now the change would be picked up by this code. Is it a good or bad thing? (good = folks would just need to check the API docs to find out; Bad = sudden changes to current code).

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: Good for me (but it needs to be AstraDBQueryResult for adhering to naming conventions). I'm changing it now.

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: I'm going to take out the `None`s in favour of boolean defaults for simplicity (and probably it's what makes the most sense here anyway).

hemidactylus: I stand corrected. The method must keep the distinction between False and None.
Astrapy complains if passing `include_similarity` at all on a non-vector search: in that case, only None is allowed.

(in other words, pending a more relaxed astrapy that defers errors to the Data API [1], a None must be allowed here).

[1] the Data API itself would not care.

libs/astradb/langchain_astradb/vectorstores.py
hemidactylus: Agree. I have updated accordingly (polymorphic return shape, overload, docstrings).

PR #115, opened 2025-02-27
Updating docs to have more vector store instantiation examples

brian-ogrady:


libs/astradb/README.md
```diff
@@ -17,19 +17,66 @@ pip install langchain-astradb
...
+vector_store_integrated = AstraDBVectorStore(
+    collection_name="astra_vector_langchain_integrated",
+    api_endpoint=ASTRA_DB_API_ENDPOINT,
+    token=ASTRA_DB_APPLICATION_TOKEN,
+    namespace=ASTRA_DB_NAMESPACE,
+    collection_vector_service_options=openai_vectorize_options,
 )
 ```
 
+Keep in mind that the `providerKey` here is not the actual value of your API key, but the [name you give your API key](https://docs.datastax.com/en/astra-db-serverless/administration/customer-keys-overview.html) when setting up your vectorize integration.
```
msmygit, line 78:
```suggestion
Keep in mind that the `providerKey` here is not the actual value of your API key, but the [name you give your API key, i.e. the unique `API_KEY_NAME`](https://docs.datastax.com/en/astra-db-serverless/integrations/embedding-providers/openai.html#add-the-openai-integration-to-collections-and-tables) when setting up your vectorize integration.
```

msmygit (changes requested):
Except for the change I requested, everything else LGTM 👍🏼

PR #119, opened 2025-02-27
Upgrade to using astrapy 2

hemidactylus:
This is a full port to astrapy 2. Not to be merged just yet as astrapy2 is still in pre-release.
The plan is for this change to start versions `0.6.x`.

Here are the main changes occurred to this codebase:

- removal of legacy "AstraDB"/"AsyncAstraDB" from constructors(/_Environment). That means `astra_db_client` and `async_astra_db_client` constructor parameters
- Remove "find_options" from constructor of `AstraDBLoader` (was deprecated for some time)
- renaming `CollectionVectorServiceOptions` ==> `VectorServiceOptions` (API unchanged). _This is just re-exported from astrapy, thus inherits the renaming there._
- Streamlined management of the "caller" overrides/additions in the _Environment classes

### DIFF
--- a/libs/astradb/langchain_astradb/utils/astradb.py
+++ b/libs/astradb/langchain_astradb/utils/astradb.py
@@ -16,13 +16,18 @@ from typing import TYPE_CHECKING, Any, Awaitable
 import langchain_core
 from astrapy import AsyncDatabase, DataAPIClient, Database
 from astrapy.admin import parse_api_endpoint
+from astrapy.authentication import (
+    EmbeddingAPIKeyHeaderProvider,
+    EmbeddingHeadersProvider,
+    StaticTokenProvider,
+    TokenProvider,
+)
 from astrapy.constants import Environment
 from astrapy.exceptions import DataAPIException
+from astrapy.info import CollectionDefinition
 
 if TYPE_CHECKING:
-    from astrapy.authentication import EmbeddingHeadersProvider, TokenProvider
-    from astrapy.db import AstraDB, AsyncAstraDB
-    from astrapy.info import CollectionDescriptor, CollectionVectorServiceOptions
+    from astrapy.info import CollectionDescriptor, VectorServiceOptions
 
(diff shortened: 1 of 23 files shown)


libs/astradb/langchain_astradb/utils/astradb.py
```diff
@@ -74,6 +80,20 @@ class SetupMode(Enum):
     OFF = 3
 
 
+def unpack_indexing_policy(
+    indexing_dict: dict[str, list[str]] | None,
+) -> tuple[str | None, list[str] | None]:
+    """{} => (None, None); {"a": "b"} => ("a", "b"); multikey => error."""
+    _indexing_dict = indexing_dict or {}
+    if _indexing_dict:
```
cbornet, line 88: I think this check is not needed

libs/astradb/langchain_astradb/utils/astradb.py
```diff
@@ -74,6 +80,20 @@ class SetupMode(Enum):
...
+        if len(_indexing_dict) != 1:
+            msg = "Unexpected indexing policy requested: " f"{indexing_dict}"
+            raise ValueError(msg)
+        keys, vals = list(zip(*_indexing_dict.items()))
+        return keys[0], vals[0]
```
cbornet, line 93:
```suggestion
        return list(_indexing_dict.items())[0]
```

libs/astradb/langchain_astradb/utils/astradb.py
hemidactylus: The check itself (the `if`) must stay, since empty/null input must => `(None, None`).

However the coercion to dict was indeed redundant. Removed - thank you!

cbornet: _indexing_dict cannot be None since there is `or {}`
And if it's empty then `len(_indexing_dict) `will not be 1.

cbornet: > However the coercion to dict was indeed redundant. Removed - thank you!

Yes, that works too.

hemidactylus:
Outdated, surpassed by #120

PR #123, opened 2025-04-24
Protect against empty-string-like ASTRA_DB_KEYSPACE environment variable

### LINKED ISSUE #122: Check if env var for keyspace is empty before using it
The pattern in astradb.py,

```
            logger.info(
                "Attempting to fetch keyspace from environment " "variable '%s'",
                KEYSPACE_ENV_VAR,
            )
            self.keyspace = os.getenv(KEYSPACE_ENV_VAR)
```

is dangerous if a workflow sets a `""` there to signal _use default_. The empty string gets pulled in, resulting in "keyspaced" requests to the Data API being actually aimed at the GeneralCommand layer (i.e. no keyspace).

Suggestion: `os.getenv(KEYSPACE_ENV_VAR)` ==> `os.getenv(KEYSPACE_ENV_VAR) or None` there.

hemidactylus:
Fixes #122 

Comes with a specific test.

### DIFF
--- a/libs/astradb/langchain_astradb/utils/astradb.py
+++ b/libs/astradb/langchain_astradb/utils/astradb.py
@@ -243,8 +243,17 @@ class _AstraDBEnvironment:
                 "Attempting to fetch keyspace from environment " "variable '%s'",
                 KEYSPACE_ENV_VAR,
             )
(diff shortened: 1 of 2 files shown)


erichare (approved):
LGTM as well!

PR #124, opened 2025-04-24
Refrain from warnings about actual lexical in actual docs if no docs

### LINKED ISSUE #121: Autodetect complains of "no lexical" when actually there's an empty collection
A misleading user warning is emitted for autodetect on an empty lexical-capable collection:

```
INFO:langchain_astradb.utils.vector_store_autodetect:vector store autodetect: inspecting 0 documents
INFO:langchain_astradb.utils.vector_store_autodetect:vector store autodetect: is_flat = False
INFO:langchain_astradb.utils.vector_store_autodetect:vector store autodetect: final_content_field = $vectorize
INFO:langchain_astradb.utils.vector_store_autodetect:vector store autodetect: lexical_in_docs = False (has_lexical = True)

UserWarning: Collection supports lexical; however, autodetect encountered documents without a $lexical. The $lexical field will be set for new documents, but the pre-existing documents might not be reached by lexical search when running hybrid search mode.
```

The silly reason of course is that [here](https://github.com/langchain-ai/langchain-datastax/blob/b840c15add12fd94505c1c62c21da0bb8bfd425d/libs/astradb/langchain_astradb/utils/vector_store_autodetect.py#L102) no attention is given to the fact that `any([]) is False`.

Fix with a None state or similar and a graceful adaptation.

hemidactylus:
Fixes #121 .

If no documents are found during autodetect, no reason to issue warnings "hey unexpected $lexical found" or "hey, caution: docs have no $lexical but it's hybrid-capable collection". These ~are~ were silly artifacts of `all([])` being True, with this PR this is no more.

### DIFF
--- a/libs/astradb/langchain_astradb/utils/vector_store_autodetect.py
+++ b/libs/astradb/langchain_astradb/utils/vector_store_autodetect.py
@@ -98,8 +98,10 @@ def _detect_documents_content_field(
     return requested_content_field
 
 
-def _detect_documents_have_lexical(documents: list[dict[str, Any]]) -> bool:
-    return any(LEXICAL_FIELD_NAME in document for document in documents)
+def _detect_documents_have_lexical(documents: list[dict[str, Any]]) -> bool | None:
+    if documents:
+        return any(LEXICAL_FIELD_NAME in document for document in documents)
+    return None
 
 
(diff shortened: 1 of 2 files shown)


erichare (approved):
LGTM!

PR #133, opened 2025-08-07
Make astradb dependencies compatible with other langchain integration

kylediaz:
Hi!

A bunch of other langchain integrations upgraded to numpy >= 2.1.0 for performance improvements. The problem is that astraDB was left out of the fun, so now it's hard to add astradb to projects because it's incompatible with other integrations!

Specifically, in python >= 3.13, most other langchain partner integrations depend on `numpy>=2.1.0` but astraDB has `^1.26.0`.

I don't see anything in the code that would imply this would break anything, and I verified that this works fine by running:
```
poetry install
poetry run python -c "
from langchain_astradb import vectorstores, cache, storage
"
```

https://github.com/langchain-ai/langchain/commit/4556b81b1d9c69eea3514b23d2ebf7bff5d471a5


kylediaz:
@hemidactylus Hi! We're looking to keep `langchain-datastax` up to date with other langchain integrations so we can utilize the latest features, but this one dependency is blocking us.

kylediaz:
Thank you!

PR #139, opened 2025-08-22
Add strict mypy checking

cbornet:


hemidactylus (approved):
LGTM, thank you! Merging
(just out of caution I have seen all CI run green locally - going ahead right now)

PR #142, opened 2025-08-26
Use uv for package management

cbornet:


.github/scripts/get_min_versions.py
```diff
@@ -1,65 +1,189 @@
...
+        package_name (str): Name of the package
+        spec_string (str): Version specification string (e.g., ">=0.2.43,<0.4.0,!=0.3.0")
+
+    Returns:
+        Optional[str]: Minimum compatible version or None if no compatible version found
+    """
+    # rewrite occurrences of ^0.0.z to 0.0.z (can be anywhere in constraint string)
+    spec_string = re.sub(r"\^0\.0\.(\d+)", r"0.0.\1", spec_string)
+    # rewrite occurrences of ^0.y.z to >=0.y.z,<0.y+1 (can be anywhere in constraint string)
+    for y in range(1, 10):
```
hemidactylus, line 64: don't these fail to catch versions such as `"^0.12.z"` ? (and likewise a few lines below)

.github/workflows/_release.yml
```diff
@@ -121,37 +136,28 @@ jobs:
...
-          poetry run pip install \
-            --extra-index-url https://test.pypi.org/simple/ \
-            "$PKG_NAME==$VERSION" || \
-          ( \
-            sleep 5 && \
-            poetry run pip install \
-              --extra-index-url https://test.pypi.org/simple/ \
-              "$PKG_NAME==$VERSION" \
-          )
+          uv venv
```
hemidactylus, line 139: It seems the "sleep 5" is gone: intentional?

.github/workflows/_test.yml
```diff
@@ -21,24 +22,19 @@ jobs:
       matrix:
         python-version:
           - "3.9"
-          - "3.10"
-          - "3.11"
-          - "3.12"
+          - "3.13"
```
hemidactylus, line 25: Isn't there value in keeping some intermediate versions? (unit tests are quick and rather cheap(

.github/scripts/get_min_versions.py
cbornet: It’s copy/pasted from langchain-community.
It is only used for langchain-core so I guess they assume they won’t have major or minor above 10.

.github/workflows/_test.yml
cbornet: It’s copy/pasted from LangChain-community.
But we can do it differently.

.github/workflows/_release.yml
```diff
@@ -13,14 +13,20 @@ on:
...
+      dangerous-nonmaster-release:
+        required: false
+        type: boolean
+        default: false
+        description: "Release from a non-master branch (danger!)"
 
 env:
-  PYTHON_VERSION: "3.12"
-  POETRY_VERSION: "1.7.1"
+  PYTHON_VERSION: "3.11"
```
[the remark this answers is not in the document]
cbornet: No idea.

.github/workflows/_release.yml
cbornet: Again this comes from LangChain-community.
I think the comment could be removed.

.github/workflows/_release.yml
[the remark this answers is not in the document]
cbornet: I changed to use Python 3.13 everywhere.

.github/workflows/_release.yml
cbornet: done

.github/workflows/_test.yml
cbornet: I added all versions.

